PR #34, opened 2021-03-26
Import ExchangeIssuance audit fixes and test updates from public repo

cgewecke:
Originally merged to [index-coop-contracts-legacy][1]

[1]: https://github.com/SetProtocol/index-coop-contracts-legacy


ncitron:
LGTM! 
One thing to note is that there is a slightly improved version of this contract here:
https://github.com/SetProtocol/index-coop-smart-contracts/blob/ncitron/reduce-dust-II/contracts/exchangeIssuance/ExchangeIssuance.sol
It has a different interface for two of the functions though so it would require changes to the frontend. Its probably fine with we don't use it (it just reduces dust for the issueExactInput functions)

ncitron (approved):

cgewecke:
@ncitron Yes, that work is also open here at #35.

(For the moment we just need a quick deploy to patch Polygon. Will definitely revisit - alex has ideas for additional work on  ExchangeIssuance)

PR #38, opened 2021-04-15
Explicitly configure localhost network gas settings

cgewecke:
Patches a problem with the way abi gas injection works when using `localhost` locally in a development flow. The gas and blockGasLimit need to be specified for the network in order for this to work.

At the moment, they default to `auto` instead of a number.

`9,500,000` is the default block limit for the Hardhat version we currently use (2.0.6)

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -37,6 +37,8 @@ const config: HardhatUserConfig = {
     },
     localhost: {
       url: "http://127.0.0.1:8545",
+      gas: 9500000,
+      blockGasLimit: 9500000,
     },
     kovan: {
       url: "https://kovan.infura.io/v3/" + process.env.INFURA_TOKEN,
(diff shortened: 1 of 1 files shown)


bweick:
I've run into issues where we end up using more gas than the limit, think we should consider bumping it to 12 million

bweick (approved):

hardhat.config.ts
```diff
@@ -37,6 +37,8 @@ const config: HardhatUserConfig = {
     },
     localhost: {
       url: "http://127.0.0.1:8545",
+      gas: 9500000,
```
cgewecke, line 40:
```suggestion
      gas: 12000000,
```

hardhat.config.ts
```diff
@@ -37,6 +37,8 @@ const config: HardhatUserConfig = {
...
+      blockGasLimit: 9500000,
```
cgewecke, line 41:
```suggestion
      blockGasLimit: 12000000,
```

PR #41, opened 2021-04-26
Decouple lib linking from hardhat & stop aliasing in tasks

cgewecke:
Targets #4


tasks/validate-dpi-params.ts
```diff
@@ -1,24 +1,24 @@
+// @ts-nocheck
```
cgewecke, line 1: This probably shouldn't be left like this long term....

PR #42, opened 2021-04-26
Supply cap / allowed contract caller issuance hook

richardliang:
Note: Set Protocol Debt Issuance Module does not have a pre redeem hook, and so the allowlist can only be applied on issuance

### DIFF
--- a/contracts/hooks/SupplyCapAllowedCallerIssuanceHook.sol
+++ b/contracts/hooks/SupplyCapAllowedCallerIssuanceHook.sol
@@ -29,7 +29,9 @@ import { ISetToken } from "../interfaces/ISetToken.sol";
  * @title SupplyCapAllowedCallerIssuanceHook
  * @author Set Protocol
  *
- * Issuance hook that checks new issuances won't push SetToken totalSupply over supply cap and checks if caller is allowed
+ * Issuance hook that checks
+ * 1) New issuances won't push SetToken totalSupply over supply cap
+ * 2) A contract address is allowed to call the module. This does not apply if caller is an EOA
  */
 contract SupplyCapAllowedCallerIssuanceHook is Ownable, IManagerIssuanceHook {
     using SafeMath for uint256;
@@ -49,7 +51,7 @@ contract SupplyCapAllowedCallerIssuanceHook is Ownable, IManagerIssuanceHook {
     // Boolean indicating if anyone can call function
     bool public anyoneCallable;
 
-    // Mapping of addresses allowed to call function
+    // Mapping of contract addresses allowed to call function
     mapping(address => bool) public callAllowList;
 
     /* ============ Constructor ============ */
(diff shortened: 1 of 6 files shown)


cgewecke:
The upfront cost problem in CI can be fixed by changing starting wallet balances to 
```ts
const TEN_MILLION_ETH = "10000000000000000000000000";
```
here:
https://github.com/SetProtocol/index-coop-smart-contracts/blob/5eedd647b9dbf3b6a10a970ecf4c2f8dc1261217/hardhat.config.ts#L59-L67

contracts/hooks/SupplyCapAllowedCallerIssuanceHook.sol
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * ONLY OWNER: Toggle ability for passed addresses to call only allowed caller functions
+     *
+     * @param _callers           Array of caller addresses to toggle status
+     * @param _statuses          Array of statuses for each caller
+     */
+    function updateCallerStatus(address[] calldata _callers, bool[] calldata _statuses) external onlyOwner {
+        require(_callers.length == _statuses.length, "Array length mismatch");
+        require(_callers.length > 0, "Array length must be > 0");
+        require(!_callers.hasDuplicate(), "Cannot duplicate callers");
```
cgewecke, line 127: We added [a set of helpers][1] to the `AddressArrayUtils` lib in set-protocol-v2 for GeneralIndexModule that make exactly these checks. They reduce this require block to...
```sol
_callers.validatePairsWithArray(_statuses);
``` 

Would it be helpful to port those over? I can open a PR for that today if so. 

[1]: https://github.com/SetProtocol/set-protocol-v2/blob/7657f9a9f77f8e639bb6650b1da5496ba2c4080b/contracts/lib/AddressArrayUtils.sol#L156-L224

contracts/hooks/SupplyCapAllowedCallerIssuanceHook.sol
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Validate if passed address is allowed to call function. If anyoneCallable is set to true, anyone can call otherwise needs to be an EOA or 
+     * approved contract address.
+     */
+    function _validateAllowedContractCaller(address _caller) internal view {
+        require(
+            _caller == tx.origin || anyoneCallable || callAllowList[_caller],
+            "Contract not permitted to call"
+        );
```
cgewecke, line 157: Just a question... in GeneralIndexModule, [the conditions][1] for `anyoneCallable` restrict the contract to EOA only. Are there reasons to be equally restrictive here?

```solidity
    /*
     * Trade must be an EOA if `anyoneTrade` has been enabled for SetToken on the module.
     */
    function _validateOnlyEOAIfUnrestricted(ISetToken _setToken) internal view {
        if(permissionInfo[_setToken].anyoneTrade) {
            require(msg.sender == tx.origin, "Caller must be EOA Address");
        }
    }
```

[1]: https://github.com/SetProtocol/set-protocol-v2/blob/7657f9a9f77f8e639bb6650b1da5496ba2c4080b/contracts/protocol/modules/GeneralIndexModule.sol#L1109-L1116

richardliang: In this case, the intention is for any one of those 3 conditions to be true to be permitted to call. Compared to the GIM, I believe even if anyone is allowed to call, it must be an EOA.

This is how I'm intending the code to work:
anyoneCallable | Is allowed contract | Is EOA | Should revert
-- | -- | -- | --
false | true | false | No
false | false | true | No
true | false | false | No
false | false | false | Yes

contracts/hooks/SupplyCapAllowedCallerIssuanceHook.sol
richardliang: Yeah this would be helpful actually

cgewecke: Those are ported in #43

cgewecke (approved):
LGTM!

PR #44, opened 2021-04-28
Update FLI viewer for multihop

richardliang:


contracts/viewers/FLIRebalanceViewer.sol
```diff
@@ -32,8 +32,12 @@ import { PreciseUnitMath } from "../lib/PreciseUnitMath.sol";
  * @title FLIRebalanceViewer
  * @author Set Protocol
  *
- * ETHFLI Rebalance viewer that returns whether a Compound oracle update should be forced before a rebalance goes through, if no
+ * FLI Rebalance viewer that returns whether a Compound oracle update should be forced before a rebalance goes through, if no
  * oracle update the type of rebalance transaction will be returned adhering to the enum specified in FlexibleLeverageStrategyAdapter
+ *
+ * CHANGELOG: 4/28/2021
+ * - Enables calculating Uniswap price with any multihop route
+ * - Lever and delever exchange data is read directly from strategy adapter and if not empty, calculate price based on encoded route
```
bweick, line 40: While we're wonder if it makes sense to add the ability to switch the `router` is pointing towards

bweick:
More general question...is the current iteration of FLI contracts compatible with UniswapV3?

PR #45, opened 2021-05-04
Add open sourced debt arb bot

richardliang:


contracts/bots/SimpleCompoundErc20FLIArb.sol
```diff
@@ -0,0 +1,547 @@
...
+import { IWETH } from "../interfaces/IWETH.sol";
+import { PreciseUnitMath } from "../lib/PreciseUnitMath.sol";
+
+/**
+ * @title SimpleCompoundErc20FLIArb
+ * @author Set Protocol
+ *
+ * Smart contract that performs creation redemption arbitrage on NAV of a FLI product against market price on Uniswap V2 or Sushiswap.
+ * Note: This contract only is compatible with FLI with 1 cToken position as collateral and 1 debt position. This contract only works
+ * with cERC20s and not cETH. Additionally, this contract is not compatible if WETH is the debt asset or if WETH is paired in the Uniswap pool
```
bweick, line 43: What's the additional complexity here? Feels like it's a decent bet we may need to build an arb bot that is compatible with WETH and cETH at some point. Is that going to be easy to do?

bweick: This info may be a bit better in a table. Just what is the asset (collateral, debt, pool) and what's excluded from being one of those assets

PR #46, opened 2021-05-18
Add escrow contracts for index sale

ncitron:


contracts/token/OtcEscrow.sol
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    modifier onlyApprovedParties() {
+        require(msg.sender == indexGov || msg.sender == beneficiary);
+        _;
+    }
+
+    /// @dev Atomically trade specified amonut of USDC for control over index in vesting contract
+    /// @dev Either counterparty may execute swap if sufficient token approval is given by recipient
+    function swap() public onlyApprovedParties {
```
bweick, line 64: So we're doing one escrow contract per beneficiary? How expensive is this function to run if it's deploying a contract during it?

contracts/token/OtcEscrow.sol
```diff
@@ -0,0 +1,91 @@
...
+pragma solidity ^0.6.10;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+import { SafeMath } from "@openzeppelin/contracts/math/SafeMath.sol";
+
+import { Vesting } from "./Vesting.sol";
+
+/*
+    Simple OTC Escrow contract to transfer index in exchange for specified USDC amount
```
bweick, line 11: There should be a * at the beginning of this line. Make sure they all lineup as well. You can find examples in the `ICManager` for formatting these javadocs

contracts/token/OtcEscrow.sol
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /// @dev Return index to Index Governance to revoke escrow deal
+    function revoke() external {
+        require(msg.sender == indexGov, "onlyIndexGovernance");
+        uint256 indexBalance = IERC20(index).balanceOf(address(this));
+        IERC20(index).safeTransfer(indexGov, indexBalance);
+    }
+
+    function revokeUsdc() external onlyApprovedParties {
```
bweick, line 87: When do we imagine this function will be used? Isn't the USDC automatically transferred out when `swap` is called?

bweick: Also seems like this should just be callable by the `beneficiary`?

contracts/token/OtcEscrow.sol
```diff
@@ -0,0 +1,91 @@
...
+        vestingEnd = _vestingEnd;
+
+        usdcAmount = _usdcAmount;
+        indexAmount = _indexAmount;
+
+        usdc = _usdcAddress;
+        index = _indexAddress;
+    }
+
+    modifier onlyApprovedParties() {
```
bweick, line 57: Not sure this modifier makes sense, never really seems like `indexGov` or `beneficiary` should be doing the same action

contracts/token/OtcEscrow.sol
bweick, line 64: Is there a reason `indexGov` would call this?

contracts/token/OtcEscrow.sol
```diff
@@ -0,0 +1,91 @@
...
+        // Transfer expected USDC from beneficiary
+        IERC20(usdc).safeTransferFrom(beneficiary, address(this), usdcAmount);
+
+        // Create Vesting contract
+        Vesting vesting = new Vesting(index, beneficiary, indexAmount, vestingStart, vestingCliff, vestingEnd);
+
+        // Transfer index to vesting contract
+        IERC20(index).safeTransfer(address(vesting), indexAmount);
+
+        // Transfer USDC to badger governance
```
bweick, line 74: `index governance`

bweick:
Couple general things: 
1) the javadocs don't really tell me what's going on. It took awhile for me to understand that there's only one beneficiary per contract for example. That's something that should definitely be noted in the contract's javadoc. Then for each function some better javadocs on who should be calling the function, when, and why would help a lot. Some of the logic in there seems unnecessary but I can't tell if maybe I'm just missing something.

2) This seems like a super expensive way for this to be done, I don't know how much the gas costs are but feels like not only are we paying a lot by having to deploy a contract for each beneficiary but the beneficiary also has to pay a ton to do the swap (since they are deploying another contract). What's the size of these checks? And how many people are we going to have to deploy these for?

contracts/token/OtcEscrow.sol
bweick: I see so IndexCoop can shoulder the costs of doing the swap if the beneficiary approves the USDC.

contracts/token/OtcEscrow.sol
ncitron: I agree this is kind of useless. It was in the original badger contracts so I had just decided to keep it in there. I figure if an investor accidentally transfers USDC to the contract it could be useful, but that seems unlikely. Maybe I should just remove this function?

bweick: I think it's fine I'd just add a javadoc that, that's what it's for

bweick: Maybe rename from `revoke` to `recover`

test/token/otcEscrow.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+      async function subject(): Promise<ContractTransaction> {
+        return await subjectOtcEscrow.connect(subjectCaller.wallet).swap();
+      }
+
+      it("should revert", async () => {
+        await expect(subject()).to.be.reverted;
+      });
+    });
+
+    context("when an inadequate amount of INDEX is sent to the escrow", async () => {
```
bweick, line 191: Instead of a silent revert maybe we can add an explicit check in the function to make sure that there's enough INDEX balance

test/token/otcEscrow.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    it("should return index when revoked", async () => {
+      const initIndexBalance = await index.balanceOf(indexGov.address);
+      await subject();
+      const finalIndexBalance = await index.balanceOf(indexGov.address);
+
+      expect(finalIndexBalance.sub(initIndexBalance)).to.eq(subjectIndexAmount);
+    });
+  });
```
bweick, line 296: Tests for `revokeUsdc` (or `recoverUsdc`)

contracts/token/OtcEscrow.sol
controtie: Does it make sense to allow anyone to call swap? The Index Coop Multisig (keys holders: Alex + Felix) will be the beneficiary. So allowing other parties to call swap will reduce the number of Multisig transactions Felix/Alex have to do.

contracts/token/OtcEscrow.sol
```diff
@@ -0,0 +1,137 @@
...
+    function recoverUsdc() external {
+        uint256 usdcBalance = IERC20(usdc).balanceOf(address(this));
+        IERC20(usdc).safeTransfer(beneficiary, usdcBalance);
+    }
+
+    /* ======= Public functions ======== */
+
+    /**
+     * Executes the OTC deal. Sends the USDC from the beneficiary to Index Governance, and
+     * locks the INDEX in the vesting contract. Can only be called by either parties of the
```
bweick, line 116: The last sentence is no longer true, are we ok with that?

test/token/otcEscrow.spec.ts
```diff
@@ -0,0 +1,340 @@
...
+        return await subjectOtcEscrow.connect(subjectCaller.wallet).swap();
+      }
+
+      it("should revert", async () => {
+        await expect(subject()).to.be.revertedWith("unauthorized");
+      });
+    });
+  });
+
+  describe("#revoke", async () => {
```
bweick, line 252: Can we get a test case for making sure IndexGov is the only one that can call this one?

test/token/otcEscrow.spec.ts
```diff
@@ -0,0 +1,340 @@
...
+      expect(await escrow.vestingCliff()).to.eq(subjectVestingCliff);
+      expect(await escrow.vestingEnd()).to.eq(subjectVestingEnd);
+      expect(await escrow.usdcAmount()).to.eq(subjectUSDCAmount);
+      expect(await escrow.indexAmount()).to.eq(subjectIndexAmount);
+      expect(await escrow.usdc()).to.eq(usdc.address);
+      expect(await escrow.index()).to.eq(index.address);
+    });
+  });
+
+  describe("#swap", async () => {
```
bweick, line 95: Let's just make a simple test case here to make sure that `beneficiary` can call as well.

hardhat.config.ts
```diff
@@ -7,6 +7,7 @@ import "@nomiclabs/hardhat-waffle";
 import "hardhat-typechain";
 import "solidity-coverage";
 import "hardhat-deploy";
+import "hardhat-gas-reporter"
```
bweick, line 10: Let's separate all of these gas-reporter updates into another PR. If we want to include these it should be a conscious decision. Don't forget to rerun `yarn install` to push the correct `yarn.lock`

test/token/otcEscrow.spec.ts
```diff
@@ -0,0 +1,374 @@
...
+      async function subject(): Promise<ContractTransaction> {
+        return await subjectOtcEscrow.connect(subjectCaller.wallet).swap();
+      }
+
+      it("should revert", async () => {
+        await expect(subject()).to.be.revertedWith("unauthorized");
+      });
+    });
+
+    context("when the caller is the investor", async () => {
```
bweick, line 251: Let's move this up above the revert cases. Generally we want to end our tests with the revert cases

contracts/token/OtcEscrow.sol
```diff
@@ -0,0 +1,139 @@
...
+    function recoverUsdc() external {
+        uint256 usdcBalance = IERC20(usdc).balanceOf(address(this));
+        IERC20(usdc).safeTransfer(beneficiary, usdcBalance);
+    }
+
+    /**
+     * Executes the OTC deal. Sends the USDC from the beneficiary to Index Governance, and
+     * locks the INDEX in the vesting contract. Can only be called by approved parties.
+     */
+    function swap() external onlyApprovedParties {
```
bweick, line 121: Can we order this function first? It's the most important one

hardhat.config.ts
```diff
@@ -53,7 +53,7 @@ const config: HardhatUserConfig = {
   typechain: {
     outDir: "typechain",
     target: "ethers-v5",
-  },
+  }
```
bweick, line 56: Let's clean up this diff

bweick (approved):

contracts/token/OtcEscrow.sol
controtie: Sorry to dig this up again, but is there any vulnerability in allowing anyone to call swap?
Right now this would require either the Index Coop MultiSig (Alex / Felix) or the investor to call swap.
I'd like to reduce the number of transactions the Index Coop Multisig must execute, and I don't think asking investors to pay the swap / contract creation cost will add friction.

Ideally I would use another wallet to call swap and get reimbursed by the Treasury Committee.

ncitron: So the concern is that if the investor has an infinite approval for usdc, someone can call swap multiple times and draw extra usdc from their wallet. I think this is not a huge concern though since the swap function will just fail since their is not enough INDEX in the contract to complete the transaction. What do you think @bweick?

bweick: Only thing I could think of here is that say (god forbid) INDEX really tanks in price someone could just deposit their INDEX here and pull USDC to the treasury. It's more of a griefing style attack since the INDEX depositor wouldn't directly benefit (unless it was the treasury itself) but it's still possible if the beneficiary does an unlimited approve.

Maybe we can remove it and when we produce instructions for the beneficiaries make sure to tell them not to do an infinite approve. But still don't like the idea of an approval hanging around out there that anyone could take advantage of.

bweick: Other option here is to add an allow list of addresses that can call this function. That way it doesn't have to be a multi-sig and it's still a somewhat trusted address

contracts/token/OtcEscrow.sol
```diff
@@ -102,7 +108,7 @@ contract OtcEscrow {
      * Executes the OTC deal. Sends the USDC from the beneficiary to Index Governance, and
      * locks the INDEX in the vesting contract. Can only be called by approved parties.
```
bweick, line 110: Let's update this javadoc to indicate it can only be called once.

test/token/otcEscrow.spec.ts
```diff
@@ -261,7 +261,8 @@ describe("OtcEscrow", () => {
       }
 
       it("should revert", async () => {
-        await expect(subject()).to.be.revertedWith("unauthorized");
+        await subject();
```
bweick, line 264: Put this `subject` in a beforeEach block, we want to make sure all the test set up doesn't happen inside the test block.

bweick (approved):

PR #50, opened 2021-05-28
Index rebalance process

cgewecke:


tasks/rebalances/calculate-new-dpi-position.ts
```diff
@@ -0,0 +1,305 @@
+// @ts-nocheck
```
bweick, line 1: This file should not be here anymore

package.json
```diff
@@ -32,6 +35,9 @@
     "precommit": "lint-staged",
     "prepare": "yarn build",
     "prepack": "if [[ \"$(basename \"$PWD\")\" == \"index-coop-contracts\" ]]; then echo \"CANNOT PUBLISH FROM THIS REPO\"; exit 1; fi;",
+    "rebalance:calc:dpi": "yarn build:lazy && BUILT=true hardhat calculate-new-dpi-position",
```
bweick, line 38: `calculate-new-dpi-position` isn't a valid script anymore

bweick: The equivalent now would be `rebalance:calc:index --index DPI`

PR #53, opened 2021-06-02
Ftc vesting contract

0xModene:


test/token/FTCVesting.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      expect(await ftcVest.treasury()).to.eq(treasury.address);
+      expect(await ftcVest.vestingAmount()).to.eq(subjectVestingAmount);
+      expect(await ftcVest.vestingBegin()).to.eq(subjectVestingStart);
+      expect(await ftcVest.vestingEnd()).to.eq(subjectVestingEnd);
+      expect(await ftcVest.vestingCliff()).to.eq(subjectVestingCliff);
+      expect(await ftcVest.lastUpdate()).to.eq(subjectVestingStart);
+    });
+  });
+
+  xdescribe("#setRecipient", async () => {
```
ncitron, line 77: This type will break your tests

test/token/FTCVesting.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    it("should set new recipient", async () => {
+      const currentRecipientAddress = await subjectFtcVesting.recipient();
+
+      expect(currentRecipientAddress).to.eq(recipient.address);
+
+      await subject();
+
+      const newRecipientAddress = subjectFtcVesting.recipient();
```
ncitron, line 115: You need to add an await here

test/token/FTCVesting.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+    let subjectFtcVesting: FTCVesting;
+
+    beforeEach(async () => {
+      const now = await getLastBlockTimestamp();
+      subjectVestingStart = now.sub(60 * 60 * 24 * 185);
+      subjectVestingCliff = now;
+      subjectVestingEnd = now.add(60 * 60 * 24 * 546);
+      subjectVestingAmount = (await index.totalSupply()).div(10);
+
+      subjectFtcVesting = await deployer.token.deployFtcVesting(
```
ncitron, line 135: After you deploy your vesting contract, you must fund it with the correct amount of INDEX

test/token/FTCVesting.spec.ts
0xModene: you mean the x? that was intentional (for now) to test things and single out the #claim test

ncitron: Oh I see, I did not realize that it was intentional

test/token/FTCVesting.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+describe("FTCVesting", () => {
+  let owner: Account;
+  let treasury: Account;
+  let recipient: Account;
+
+  let deployer: DeployHelper;
+  let index: IndexToken;
+
+  before(async () => {
+    owner = await getRandomAccount();
```
ncitron, line 27: You can use array destructuring here to clean this up a bit like this:
```ts
[
    owner,
    treasury,
    recipient
] = await getAcounts();
```

test/token/FTCVesting.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+    });
+
+    async function subject(): Promise<ContractTransaction> {
+      return subjectFtcVesting.connect(recipient.wallet).setRecipient(newRecipient.address);
+    }
+
+    it("should set new recipient address", async () => {
+      const currentRecipientAddress = await subjectFtcVesting.recipient();
+
+      expect(currentRecipientAddress).to.eq(recipient.address);
```
ncitron, line 111: No need to have this expect here. This is covered by the prior test.

test/token/FTCVesting.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+      );
+
+      await index.transfer(subjectFtcVesting.address, ether(1000));
+    });
+
+    async function subject(): Promise<ContractTransaction> {
+      return await subjectFtcVesting.connect(recipient.wallet).claim();
+    }
+
+    it("should make a claim", async () => {
```
ncitron, line 196: Look into using `increaseTimeAsync` so you can do some more specific tests (like ensuring that 100% of the rewards have been distributed after the full time has elapsed)

test/token/FTCVesting.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+    async function subject(): Promise<ContractTransaction> {
+      return await subjectFtcVesting.connect(treasury.wallet).clawback();
+    }
+
+    it("should clawback funds", async () => {
+      const unclaimedAmount = await index.connect(treasury.wallet).balanceOf(treasury.address);
+      await subject();
+
+      const claimedAmount = await index.connect(treasury.wallet).balanceOf(treasury.address);
+      const hasClaimed = claimedAmount.gt(unclaimedAmount);
```
ncitron, line 247: Can you be more specific here? We should ensure that 100% of the funds have been transferred from the vesting contract to the treasury after the clawback rather than just ensuring some amount has been moved. Also, can we use some more descriptive variables for `unclaimedaAmount` and `claimedAmount`. Maybe `initTreasuryBalance` and `finalTreasuryBalance`

test/token/FTCVesting.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+
+      const hasClaimed = claimedAmount.gt(unclaimedAmount);
+      const hasLowerBalance = preClaimBalance.gt(postClaimBalance);
+
+      expect(hasClaimed).to.be.true;
+      expect(hasLowerBalance).to.be.true;
+    });
+  });
+
+  describe("#clawback", async () => {
```
ncitron, line 211: For every permissioned function you write, you want to add a test to ensure that it reverts if the wrong person calls it. Take a look at how its done using `context` here:
https://github.com/SetProtocol/index-coop-smart-contracts/blob/master/test/token/otcEscrow.spec.ts#L313

utils/deploys/deployToken.ts
```diff
@@ -7,6 +7,8 @@ import { IndexToken__factory } from "../../typechain/factories/IndexToken__facto
 import { MerkleDistributor__factory } from "../../typechain/factories/MerkleDistributor__factory";
 import { Vesting__factory } from "../../typechain/factories/Vesting__factory";
 import { OtcEscrow__factory } from "../../typechain/factories/OtcEscrow__factory";
+import { FTCVesting__factory } from "@typechain/factories/FTCVesting__factory";
```
ncitron, line 10: Can you use the `../../typechain` formatting here  so that it falls in line with how the rest of the repo formats it (I agree that your way looks better though)

utils/deploys/deployToken.ts
```diff
@@ -7,6 +7,8 @@ import { IndexToken__factory } from "../../typechain/factories/IndexToken__facto
...
+import { FTCVesting } from "@typechain/FTCVesting";
```
ncitron, line 11: Can you export FTCVesting in `utils/contracts.ts` so that you can import it with the rest of the contract types on line 4

0xModene:
@ncitron I think I have addressed all of your concerns

test/token/FTCVesting.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+      );
+
+      await index.transfer(subjectFtcVesting.address, ether(1000));
+    });
+
+    async function subject(): Promise<ContractTransaction> {
+      return await subjectFtcVesting.connect(recipient.wallet).claim();
+    }
+
+    it("should make a claim", async () => {
```
ncitron, line 193: Can we split this into two tests? One for a partial claim and one for a full claim. Typically you only want to call subject() once per test.

contracts/token/FTCVesting.sol
```diff
@@ -0,0 +1,92 @@
...
+    constructor(
+        address index_,
+        address recipient_,
+        address treasury_,
+        uint256 vestingAmount_,
+        uint256 vestingBegin_,
+        uint256 vestingCliff_,
+        uint256 vestingEnd_
+    ) public {
+        require(vestingCliff_ >= vestingBegin_, "FTCVester.constructor: cliff is too early");
```
richardliang, line 34: What's the reasoning for removing this check in the Vesting contract?
`require(vestingBegin_ >= block.timestamp, "TreasuryVester.constructor: vesting begin too early");`

Is it retroactively vesting for contributors? Add a comment here noting why we removed the require for clarity.

richardliang (approved):
Checked against diff of Vesting.sol. Had a question on why the require for `vestingBegin_` but LGTM if that's resolved

contracts/token/FTCVesting.sol
0xModene: This was removed to allow retroactive vesting. We are currently approaching the cliff date on a few contributors, so this was removed to keep the accounting data straight with the hire date, vesting begin, etc.

PR #56, opened 2021-06-15
Add ExchangeIssuanceV2 contract

cgewecke:
PR adds a simplified variant of the ExchangeIssuance contract which has all native currency logic removed. 

The rationale for doing this is two-part:
+ non-ethereum chains have non-ETH native currencies. `ExchangeIssuance`'s WETH unwrapping logic becomes WMATIC unwrapping (or equivalent) when executed on a side chain. 
+ WETH is still the most liquid component in AMM pairs on the side chain we're targeting (Polygon). Substituting a native currency wrapper like WMATIC for WETH would impair trade performance. 

Contract changes here are:

**Method removals**
+ [receive][1]
+ [issueSetForExactETH][2]
+ [issueExactSetFromETH][3]
+ [redeemExactSetForETH][4]

[1]: https://github.com/SetProtocol/index-coop-smart-contracts/blob/7ef60ef3f007de8dd40a9cfcd0c0dd1e0a3b693c/contracts/exchangeIssuance/ExchangeIssuance.sol#L145-L148
[2]: https://github.com/SetProtocol/index-coop-smart-contracts/blob/7ef60ef3f007de8dd40a9cfcd0c0dd1e0a3b693c/contracts/exchangeIssuance/ExchangeIssuance.sol#L224-L242
[3]: https://github.com/SetProtocol/index-coop-smart-contracts/blob/7ef60ef3f007de8dd40a9cfcd0c0dd1e0a3b693c/contracts/exchangeIssuance/ExchangeIssuance.sol#L297-L323
[4]: https://github.com/SetProtocol/index-coop-smart-contracts/blob/7ef60ef3f007de8dd40a9cfcd0c0dd1e0a3b693c/contracts/exchangeIssuance/ExchangeIssuance.sol#L384-L414

**Logic changes**
 
The WETH redemption flow `issueExactSetFromToken` has been changed from "unwrap WETH and sendValue" to "WETH.transfer"

**Old**
https://github.com/SetProtocol/index-coop-smart-contracts/blob/ecef4a21cc6161c3651b72bc6da6ffa787af2dd7/contracts/exchangeIssuance/ExchangeIssuance.sol#L277-L281

**New**
https://github.com/SetProtocol/index-coop-smart-contracts/blob/7ef60ef3f007de8dd40a9cfcd0c0dd1e0a3b693c/contracts/exchangeIssuance/ExchangeIssuanceV2.sol#L243-L246

The only tests with substantive changes (apart from those removed) to accommodate the new behavior are in 

[exchangeIssuanceV2.spec.ts#L754-L798][5]

[5]: https://github.com/SetProtocol/index-coop-smart-contracts/blob/7ef60ef3f007de8dd40a9cfcd0c0dd1e0a3b693c/test/exchangeIssuance/exchangeIssuanceV2.spec.ts#L754-L798

### DIFF
--- a/contracts/exchangeIssuance/ExchangeIssuanceV2.sol
+++ b/contracts/exchangeIssuance/ExchangeIssuanceV2.sol
@@ -1,0 +1,808 @@
+/*
+    Copyright 2021 Index Cooperative
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+    SPDX-License-Identifier: Apache License, Version 2.0
+*/
+pragma solidity 0.6.10;
+pragma experimental ABIEncoderV2;
+
+import { Address } from "@openzeppelin/contracts/utils/Address.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { IUniswapV2Factory } from "@uniswap/v2-core/contracts/interfaces/IUniswapV2Factory.sol";
+import { IUniswapV2Router02 } from "@uniswap/v2-periphery/contracts/interfaces/IUniswapV2Router02.sol";
+import { Math } from "@openzeppelin/contracts/math/Math.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+import { SafeMath } from "@openzeppelin/contracts/math/SafeMath.sol";
+import { ReentrancyGuard } from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+
+import { IBasicIssuanceModule } from "../interfaces/IBasicIssuanceModule.sol";
+import { IController } from "../interfaces/IController.sol";
+import { ISetToken } from "../interfaces/ISetToken.sol";
+import { IWETH } from "../interfaces/IWETH.sol";
+import { PreciseUnitMath } from "../lib/PreciseUnitMath.sol";
+import { UniSushiV2Library } from "../../external/contracts/UniSushiV2Library.sol";
+
+/**
+ * @title ExchangeIssuance
+ * @author Index Coop
+ *
+ * Contract for issuing and redeeming any SetToken using ETH or an ERC20 as the paying/receiving currency.
+ * All swaps are done using the best price found on Uniswap or Sushiswap.
+ *
+ */
+contract ExchangeIssuanceV2 is ReentrancyGuard {
+
+    using Address for address payable;
+    using SafeMath for uint256;
+    using PreciseUnitMath for uint256;
+    using SafeERC20 for IERC20;
+    using SafeERC20 for ISetToken;
+
+    /* ============ Enums ============ */
+
+    enum Exchange { Uniswap, Sushiswap, None }
+
+    /* ============ Constants ============= */
+
+    uint256 constant private MAX_UINT96 = 2**96 - 1;
+    address constant public ETH_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+
+    /* ============ State Variables ============ */
+
+    address public WETH;
+    IUniswapV2Router02 public uniRouter;
+    IUniswapV2Router02 public sushiRouter;
+
+    address public immutable uniFactory;
+    address public immutable sushiFactory;
+
+    IController public immutable setController;
+    IBasicIssuanceModule public immutable basicIssuanceModule;
+
+    /* ============ Events ============ */
+
+    event ExchangeIssue(
+        address indexed _recipient,     // The recipient address of the issued SetTokens
+        ISetToken indexed _setToken,    // The issued SetToken
+        IERC20 indexed _inputToken,     // The address of the input asset(ERC20/ETH) used to issue the SetTokens
+        uint256 _amountInputToken,      // The amount of input tokens used for issuance
+        uint256 _amountSetIssued        // The amount of SetTokens received by the recipient
+    );
+
+    event ExchangeRedeem(
+        address indexed _recipient,     // The recipient address which redeemed the SetTokens
+        ISetToken indexed _setToken,    // The redeemed SetToken
+        IERC20 indexed _outputToken,    // The address of output asset(ERC20/ETH) received by the recipient
+        uint256 _amountSetRedeemed,     // The amount of SetTokens redeemed for output tokens
+        uint256 _amountOutputToken      // The amount of output tokens received by the recipient
+    );
+
+    event Refund(
+        address indexed _recipient,     // The recipient address which redeemed the SetTokens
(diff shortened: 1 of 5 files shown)


cgewecke:
This is a unified diff of ExchangeIssuance vs. ExchangeIssuanceV2

```diff
33d32
< 
42c41
< contract ExchangeIssuance is ReentrancyGuard {
---
> contract ExchangeIssuanceV2 is ReentrancyGuard {
145,149d143
<     receive() external payable {
<         // required for weth.withdraw() to work properly
<         require(msg.sender == WETH, "ExchangeIssuance: Direct deposits not allowed");
<     }
< 
217,244d210
<      * Issues SetTokens for an exact amount of input ether.
<      *
<      * @param _setToken         Address of the SetToken to be issued
<      * @param _minSetReceive    Minimum amount of SetTokens to receive. Prevents unnecessary slippage.
<      *
<      * @return setTokenAmount   Amount of SetTokens issued to the caller
<      */
<     function issueSetForExactETH(
<         ISetToken _setToken,
<         uint256 _minSetReceive
<     )
<         isSetToken(_setToken)
<         external
<         payable
<         nonReentrant
<         returns(uint256)
<     {
<         require(msg.value > 0, "ExchangeIssuance: INVALID INPUTS");
< 
<         IWETH(WETH).deposit{value: msg.value}();
< 
<         uint256 setTokenAmount = _issueSetForExactWETH(_setToken, _minSetReceive, msg.value);
< 
<         emit ExchangeIssue(msg.sender, _setToken, IERC20(ETH_ADDRESS), msg.value, setTokenAmount);
<         return setTokenAmount;
<     }
< 
<     /**
279,280c245
<             IWETH(WETH).withdraw(amountEthReturn);
<             (payable(msg.sender)).sendValue(amountEthReturn);
---
>             IERC20(WETH).safeTransfer(msg.sender,  amountEthReturn);
289,325d253
<     * Issues an exact amount of SetTokens using a given amount of ether.
<     * The excess ether is returned back.
<     *
<     * @param _setToken          Address of the SetToken being issued
<     * @param _amountSetToken    Amount of SetTokens to issue
<     *
<     * @return amountEthReturn   Amount of ether returned to the caller
<     */
<     function issueExactSetFromETH(
<         ISetToken _setToken,
<         uint256 _amountSetToken
<     )
<         isSetToken(_setToken)
<         external
<         payable
<         nonReentrant
<         returns (uint256)
<     {
<         require(msg.value > 0 && _amountSetToken > 0, "ExchangeIssuance: INVALID INPUTS");
< 
<         IWETH(WETH).deposit{value: msg.value}();
< 
<         uint256 amountEth = _issueExactSetFromWETH(_setToken, _amountSetToken, msg.value);
< 
<         uint256 amountEthReturn = msg.value.sub(amountEth);
< 
<         if (amountEthReturn > 0) {
<             IWETH(WETH).withdraw(amountEthReturn);
<             (payable(msg.sender)).sendValue(amountEthReturn);
<         }
< 
<         emit Refund(msg.sender, amountEthReturn);
<         emit ExchangeIssue(msg.sender, _setToken, IERC20(ETH_ADDRESS), amountEth, _amountSetToken);
<         return amountEthReturn;
<     }
< 
<     /**
374,414d301
<     /**
<      * Redeems an exact amount of SetTokens for ETH.
<      * The SetToken must be approved by the sender to this contract.
<      *
<      * @param _setToken             Address of the SetToken to be redeemed
<      * @param _amountSetToken       Amount of SetTokens to redeem
<      * @param _minEthOut            Minimum amount of ETH to receive
<      *
<      * @return amountEthOut         Amount of ether sent to the caller
<      */
<     function redeemExactSetForETH(
<         ISetToken _setToken,
<         uint256 _amountSetToken,
<         uint256 _minEthOut
<     )
<         isSetToken(_setToken)
<         external
<         nonReentrant
<         returns (uint256)
<     {
<         require(_amountSetToken > 0, "ExchangeIssuance: INVALID INPUTS");
< 
<         address[] memory components = _setToken.getComponents();
<         (
<             uint256 totalEth,
<             uint256[] memory amountComponents,
<             Exchange[] memory exchanges
<         ) =  _getAmountETHForRedemption(_setToken, components, _amountSetToken);
< 
<         require(totalEth > _minEthOut, "ExchangeIssuance: INSUFFICIENT_OUTPUT_AMOUNT");
< 
<         _redeemExactSet(_setToken, _amountSetToken);
< 
<         uint256 amountEthOut = _liquidateComponentsForWETH(components, amountComponents, exchanges);
< 
<         IWETH(WETH).withdraw(amountEthOut);
<         (payable(msg.sender)).sendValue(amountEthOut);
< 
<         emit ExchangeRedeem(msg.sender, _setToken, IERC20(ETH_ADDRESS), _amountSetToken, amountEthOut);
<         return amountEthOut;
<     }
```

contracts/exchangeIssuance/ExchangeIssuanceV2.sol
```diff
@@ -0,0 +1,808 @@
...
+     * @param _tokenA       The address of tokenA
+     * @param _tokenB       The address of tokenB
+     *
+     * @return              The max amount of tokens that can be received across both exchanges
+     * @return              The Exchange on which maximum amount of token can be received
+     * @return              The pair address of the uniswap/sushiswap pool containing _tokenA and _tokenB
+     */
+    function _getMaxTokenForExactToken(uint256 _amountIn, address _tokenA, address _tokenB) internal view returns (uint256, Exchange, address) {
+        if (_tokenA == _tokenB) {
+            return (_amountIn, Exchange.None, ETH_ADDRESS);
```
asoong, line 761: Can someone clarify how the ETH address is being used? Is this just an arbitrary address when the currency token is in the Set?

asoong (approved):
LGTM, I think this should start belonging in set protocol contracts. It was previously committed here because the major contributors were community members but we will be expanding upon this in the set protocol contracts repo with multiple exchange types.

contracts/exchangeIssuance/ExchangeIssuanceV2.sol
cgewecke: I'm going to ask @ncitron about this quickly. It looks like there's a single place the address could be read:
https://github.com/SetProtocol/index-coop-smart-contracts/blob/7ef60ef3f007de8dd40a9cfcd0c0dd1e0a3b693c/contracts/exchangeIssuance/ExchangeIssuanceV2.sol#L640

cgewecke: On slack Noah says: 
> It was originally added so that when we emitted events for the ETH based functions, we could use it as the input/output token value to denote that we issued/redeemed with ETH rather than WETH

> [uncertain] why that got added in the early return for _getMaxTokenForExactToken function. I see no instances when tokenA would even equal tokenB

> We didn't have that check in the original version:

The logic was introduced in: https://github.com/SetProtocol/index-coop-contracts-legacy/pull/13/files

We may have to revisit this and re-deploy with a patch but think this looks ok for staging testing...

cgewecke: Sachin confirmed it's ok to ignore this and suggested ETH_ADDRESS could be replaced with ZERO address. 

From slack:

> Regarding the ETH_ADDRESS , when the tokens A and B are same in _getMinTokenForExactToken, we do not need to swap. Which might happen when the input token, say ETH, which is used to acquire other SetToken components, IS one of the SetToken components. So, we need not acquire it from an exchange, and can just use the input tokens provided.

> In that case the return values from `_getMinTokenForExactToken` contains, `Exchange.None`, signifying no exchange, and the pair address should have been ZERO address, but it was a mistake by me, to return ETH_ADDRESS as the pair address there.

PR #59, opened 2021-06-19
Add trade routing to FlexibleLeverageStrategyAdapter

ncitron:
ITIP: https://github.com/SetProtocol/ITIPS/blob/brian/fli-trade-routing/ITIPS/ITIP-001.md


contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -938,11 +1062,14 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
         uint256 _chunkRebalanceNotional,
         uint256 _totalRebalanceNotional,
-        uint256 _newLeverageRatio
+        uint256 _newLeverageRatio,
+        string memory _exchangeName
     )
         internal
     {
-        lastTradeTimestamp = block.timestamp;
+        globalLastTradeTimestamp = block.timestamp;
```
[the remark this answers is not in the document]
richardliang: Could be cleaner if we consolidate these 2 lines into another internal function _updateLastTradeTimestamp function since its used across _updateRebalanceState, _updateIterateState, _updateRipcordState

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -270,11 +281,17 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
-        LeverageInfo memory leverageInfo = _getAndValidateLeveragedInfo(execution.slippageTolerance, execution.twapMaxTradeSize);
+     function rebalance(string memory _exchangeName) external onlyEOA onlyAllowedCaller(msg.sender) {
+        LeverageInfo memory leverageInfo = _getAndValidateLeveragedInfo(
+            execution.slippageTolerance,
+            exchangeSettings[_exchangeName].twapMaxTradeSize,
+            _exchangeName
+        );
 
-        _validateNormalRebalance(leverageInfo, methodology.rebalanceInterval);
+        _validateNormalRebalance(leverageInfo, methodology.rebalanceInterval, globalLastTradeTimestamp);
```
richardliang, line 294: Comment here why we're passing in global last trade timestamp

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -297,11 +314,17 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
-        LeverageInfo memory leverageInfo = _getAndValidateLeveragedInfo(execution.slippageTolerance, execution.twapMaxTradeSize);
+    function iterateRebalance(string memory _exchangeName) external onlyEOA onlyAllowedCaller(msg.sender) {
+        LeverageInfo memory leverageInfo = _getAndValidateLeveragedInfo(
+            execution.slippageTolerance,
+            exchangeSettings[_exchangeName].twapMaxTradeSize,
+            _exchangeName
+        );
 
-        _validateNormalRebalance(leverageInfo, execution.twapCooldownPeriod);
+        _validateNormalRebalance(leverageInfo, execution.twapCooldownPeriod, exchangeSettings[_exchangeName].exchangeLastTradeTimestamp);
```
richardliang, line 327: Comment here why we're passing in the exchange specific timestamp vs global

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -327,20 +350,23 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
+    function ripcord(string memory _exchangeName) external onlyEOA {
         LeverageInfo memory leverageInfo = _getAndValidateLeveragedInfo(
             incentive.incentivizedSlippageTolerance, 
-            incentive.incentivizedTwapMaxTradeSize
+            exchangeSettings[_exchangeName].incentivizedTwapMaxTradeSize,
+            _exchangeName
         );
 
-        _validateRipcord(leverageInfo);
+        _validateRipcord(leverageInfo, exchangeSettings[_exchangeName].exchangeLastTradeTimestamp);
```
richardliang, line 363: Comment here why we're passing in exchange specific timestamp

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -417,12 +449,8 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
             execution.unutilizedLeveragePercentage,
-            execution.twapMaxTradeSize,
             execution.twapCooldownPeriod,
-            execution.slippageTolerance,
-            execution.exchangeName,
-            execution.leverExchangeData,
-            execution.deleverExchangeData
+            execution.slippageTolerance
         );
     }
```
richardliang, line 456: Add event for ExchangeSettingsUpdated

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -469,6 +554,36 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
+     * Calculates the total notional rebalance size.
+     *
+     * @return uint256      total notional rebalance size
+     */
+    function getTotalRebalanceNotional()  external view returns(uint256) {
+
+        uint256 newLeverageRatio;
+        uint256 currentLeverageRatio = getCurrentLeverageRatio();
+
+        if (currentLeverageRatio > incentive.incentivizedLeverageRatio) {
```
richardliang, line 567: I can guess why you're breaking it out this way instead of using `_calculateNewLeverageRatio`, but we need some clarifying comments on why and lay out the scenarios we're trying to cover here. I.e. there could be situations where we're in ripcord territory, but there is a twapLeverageRatio set from a previous iterateRebalance, which would overridden by a ripcord call. Therefore, we want to check if it's above the incentivized leverage ratio directly

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -469,6 +554,36 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
+        } else if (twapLeverageRatio > 0) {
+            if (_isAdvantageousTWAP(currentLeverageRatio)) {
+                return 0;
+            }
+            newLeverageRatio = twapLeverageRatio;
+        } else {
+            newLeverageRatio = _calculateNewLeverageRatio(currentLeverageRatio);
+        }
+
+        bool isLever = newLeverageRatio > currentLeverageRatio;
```
richardliang, line 578: We don't need to create this variable if we directly do `newLeverageRatio > currentLeverageRatio` in line 581

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -441,10 +469,67 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
+     */
+    function addEnabledExchange(
+        string memory _exchangeName,
+        ExchangeSettings memory _exchangeSettings
+    )
+        external 
+        onlyOperator 
+        noRebalanceInProgress
+    {
+        require(exchangeSettings[_exchangeName].twapMaxTradeSize == 0, "Exchange already enabled");
```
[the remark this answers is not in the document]
richardliang: We can create an internal function similar to _validateSettings, maybe _validateExchangeSettings that will contain these 2 validations. 

I know I removed it in the spec, but we can also add back the validation of twapMaxTradeSize <= incentivizedTwapMaxTradeSize

So 3 validations in total

test/integration/fliIntegration.spec.ts
```diff
@@ -4,11 +4,11 @@ import { defaultAbiCoder } from "ethers/lib/utils";
```
richardliang, line 4: Let's write some integration tests where we rebalance using multiple exchanges when it's above the max leverage ratio, and also iterate rebalance using multiple exchanges

test/adapters/flexibleLeverageStrategyAdapter.spec.ts
```diff
@@ -838,10 +879,12 @@ describe("FlexibleLeverageStrategyAdapter", () => {
   describe("#rebalance", async () => {
```
richardliang, line 879: Can we write a test where theres 2 exchanges, we call rebalance on the 1st exchange, and then call rebalance on the second exchange right after, so the 2nd transaction would revert since rebalance interval has not elapsed

test/adapters/flexibleLeverageStrategyAdapter.spec.ts
```diff
@@ -1542,11 +1629,13 @@ describe("FlexibleLeverageStrategyAdapter", () => {
   describe("#iterateRebalance", async () => {
```
richardliang, line 1629: Can we write a test with 2 exchanges, we call iterateRebalance on both exchanges immediately after one another, where the global and exchange timestamps are updated correctly

test/adapters/flexibleLeverageStrategyAdapter.spec.ts
richardliang: Then also a passing case where we're above or below the bounds and timestamps are updated correctly

test/adapters/flexibleLeverageStrategyAdapter.spec.ts
```diff
@@ -2034,10 +2153,12 @@ describe("FlexibleLeverageStrategyAdapter", () => {
   describe("#ripcord", async () => {
```
richardliang, line 2153: Can we write a test with 2 exchanges, we call ripcord on both exchanges immediately after one another, where the timestamps are updated correctly

richardliang (commented):
Nice! We can improve on some comments, add some more validations, and can write more tests to ensure state is consistent especially with the introduction of the new exchange specific timestamp variable

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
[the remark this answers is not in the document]
ncitron: Is it necessary that the normal max trade size is less than the incentivized one? I imagine there could be an exchange that we may want to disable for ripcords in the future.

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -469,6 +598,43 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
         return _calculateCurrentLeverageRatio(currentLeverageInfo.collateralValue, currentLeverageInfo.borrowValue);
     }
 
+    /**
+     * Calculates the total notional rebalance size. Note: this function does not take into account timestamps, so it may return a nonzero value
+     * even when shouldRebalance would return ShouldRebalance.NONE for all exchanges (since minimum delays have not elapsed)
+     *
+     * @return uint256      total notional rebalance size
+     */
+    function getTotalRebalanceNotional()  external view returns(uint256) {
```
richardliang, line 607: For getTotalRebalanceNotional(), think we want to return a bool and an uint. The bool being whether its lever or delever, and the uint is the quantity in whatever asset is being sold (sent to the exchange). So for delever, it is notional quantity in WETH (sold for USDC) and for lever it is notional USDC (sold for WETH). This makes the UX much better for keeper bots. Currently, we’re only returning the notional in collateral asset both directions

For lever, we can convert the collateral notional into borrow units using this internal function _calculateBorrowUnits

test/adapters/flexibleLeverageStrategyAdapter.spec.ts
```diff
@@ -1989,9 +2199,60 @@ describe("FlexibleLeverageStrategyAdapter", () => {
...
+      describe("when in a twap rebalance and under target leverage ratio", async () => {
+        it("should set the global and exchange timestamps correctly", async () => {
+          await subject();
+          const timestamp1 = await getLastBlockTimestamp();
+
+          subjectExchangeToUse = exchangeName2;
+          await subject();
+          const timestamp2 = await getLastBlockTimestamp();
+
+          expect(await flexibleLeverageStrategyAdapter.globalLastTradeTimestamp()).to.eq(timestamp2);
```
richardliang, line 2246: Nice!

contracts/mocks/StringArrayUtilsMock.sol
```diff
@@ -0,0 +1,41 @@
+/*
+    Copyright 2021 Set Labs Inc.
```
richardliang, line 2: Spacing between paragraphs seem off

richardliang (commented):
Getting really close here. We'd want to change the getTotalRebalanceNotional() so it is more useful for keepers / other contracts

Also missing a few lines in Coveralls https://coveralls.io/builds/40761194/source?filename=contracts%2Fadapters%2FFlexibleLeverageStrategyAdapter.sol

test/adapters/flexibleLeverageStrategyAdapter.spec.ts
```diff
@@ -1501,6 +1620,65 @@ describe("FlexibleLeverageStrategyAdapter", () => {
...
+            await subject();
+            subjectExchangeToUse = exchangeName2;
+          });
+
+          it("should revert", async () => {
+            await expect(subject()).to.be.revertedWith("Cooldown not elapsed or not valid leverage ratio");
+          });
+        });
+      });
+
```
richardliang, line 1681: One more test case just for our sanity. When leverage ratio is within bounds and rebalance is called before rebalance interval has elapsed

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -184,30 +217,36 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
-    uint256 public lastTradeTimestamp;              // Last rebalance timestamp. Must be past rebalance interval to rebalance
+    ContractSettings internal strategy;                             // Struct of contracts used in the strategy (SetToken, price oracles, leverage module etc)
+    MethodologySettings internal methodology;                       // Struct containing methodology parameters
+    ExecutionSettings internal execution;                           // Struct containing execution parameters
+    mapping(string => ExchangeSettings) internal exchangeSettings;  // Mapping from exchange name to exchange settings
+    IncentiveSettings internal incentive;                           // Struct containing incentive parameters for ripcord
+    string[] public enabledExchanges;                               // Array containing enabled exchanges
+    uint256 public twapLeverageRatio;                               // Stored leverage ratio to keep track of target between TWAP rebalances
+    uint256 public globalLastTradeTimestamp;                        // Last rebalance timestamp. Must be past rebalance interval to rebalance
```
richardliang, line 227: More specificity on that this, current timestamp must be greater than this variable by the rebalance interval to rebalance

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -469,6 +612,39 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
         return _calculateCurrentLeverageRatio(currentLeverageInfo.collateralValue, currentLeverageInfo.borrowValue);
     }
 
+    /**
+     * Calculates the total notional rebalance size. Note: this function does not take into account timestamps, so it may return a nonzero value
```
richardliang, line 616: Add a bit more context. `Calculates the total notional rebalance size that can be used by external contracts or keeper bots to determine optimal exchange to rebalance on`

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -526,13 +702,23 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
         return _shouldRebalance(currentLeverageRatio, _customMinLeverageRatio, _customMaxLeverageRatio);
     }
 
+    /**
+     * Gets the list of enabled exchanges
+     */
+    function getEnabledExchanges() external view returns (string[] memory) {
```
richardliang, line 708: We can probably just make the state variable public and remove this function

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -975,6 +1174,14 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
         }
     }
 
+    /**
+     * Update globalLastTradeTimestamp and exchangeLastTradeTimestamp values
```
richardliang, line 1178: Add a bit more context on why we always need to update both. i.e. our assumption that global must always be >= exchange specific timestamp

richardliang (commented):
Round of review for comments. Overall, looks pretty good to me

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
ncitron: This is here mostly to make testing that exchanges have been removed properly easier. Otherwise, you have to check if getting the nth element reverts (since there is no way to check the length of an array exposed via a public method)

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -748,20 +938,25 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
-            "TWAP max trade size must be less than incentivized TWAP max trade size"
-        );
     }
 
+    /**
+     * Validate an ExchangeSettings struct when adding or updating an exchange. Does not validate that twapMaxTradeSize < incentivizedMaxTradeSize since
+     * it may be useful to disable exchanges for ripcord by setting incentivizedMaxTradeSize to 0.
+     */
+     function _validateExchangeSettings(ExchangeSettings memory _settings) internal pure {
+         require(_settings.exchangeLastTradeTimestamp == 0, "Last trade timestamp must be 0");
```
bweick, line 948: This implies we will always reset the timestamp when we update an exchange. Is that what we want to do?

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -469,6 +612,39 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
+        } else {
+            newLeverageRatio = _calculateNewLeverageRatio(currentLeverageRatio);
+        }
+
+        ActionInfo memory actionInfo = _createActionInfo();
+
+        bool isLever = newLeverageRatio > currentLeverageRatio;
+        uint256 leverageRatioDifference = isLever ? newLeverageRatio.sub(currentLeverageRatio) : currentLeverageRatio.sub(newLeverageRatio);
+
+        uint256 collateralRebalanceUnits = leverageRatioDifference.preciseDiv(currentLeverageRatio).preciseMul(actionInfo.collateralBalance);
```
bweick, line 643: This should be `notional` not in `units` so think this name is a misnomer

bweick: Feels like there may be some duplicate logic here and in `_calculateChunkRebalanceNotional `. Which brings up another point do we want to return the`chunkRebalanceNotional` or the `totalRebalanceNotional`?

richardliang: It's way more complicated doing `chunkRebalanceNotional` because we'd need to pass in the `exchangeName` and fetch the max trade size there. Instead, we're planning to have the FLI rebalance viewer logic take care of the chunking logic and passing the chunk notional into the exchange get quote functions

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
richardliang: Yeah think we can make it 100% clear by individually assigning the variables as you suggested above. It won't cause any weird state if exchange timestamp resets to 0 since the daily rebalance will read the global timestamp and we don't allow exchange setting updates when a rebalance is in progress (so won't affect TWAP cooldown period either)

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -435,14 +505,87 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
+        ExchangeSettings memory _exchangeSettings
+    )
+        external 
+        onlyOperator 
+        noRebalanceInProgress
+    {
+        require(exchangeSettings[_exchangeName].twapMaxTradeSize == 0, "Exchange already enabled");
+        _validateExchangeSettings(_exchangeSettings);
+        
+        exchangeSettings[_exchangeName] = _exchangeSettings;
```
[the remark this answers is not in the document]
richardliang: It won't cause any weird state, but think we should do this to make the code 100% clear

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
richardliang: gotcha

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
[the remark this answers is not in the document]
ncitron: Should we explicitly set the timestamp to 0 when adding a new exchange? I think it would make the most sense since we might want to use an exchange immediately after adding it

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
ncitron: I think we can do it either way. It all just depends on whether we want the logic for calculating chunks to go in the viewer contract or here. Any agreement on which way to go with?

ncitron: I think we can do it either way. It all just depends on whether we want the logic for calculating chunks to go in the viewer contract or here. Any agreement on which way to go with?

bweick: My feeling is that we should pass in an array of exchange names that returns an array of chunk sizes. I think that's a more practical solution and reduces any logic duplication across contracts.

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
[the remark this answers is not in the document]
richardliang: yeah that makes sense for a new exchange. For changing existing exchanges, imagine it would just use the existing timestamp.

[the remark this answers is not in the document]
bweick: I'm fine with a new exchange starting at 0. Where it gets iffy is when we are updating an exchange do we reset the timestamp? If we do then I don't think we want to allow exchange updates to happen during a rebalance in case we accidentally trade during a cool down period.

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -435,14 +505,89 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
+        require(exchangeSettings[_exchangeName].twapMaxTradeSize != 0, "Exchange not enabled");
+
+        delete exchangeSettings[_exchangeName];
+        enabledExchanges.removeStorage(_exchangeName);
+
+        emit ExchangeRemoved(_exchangeName);
+    }
+
+    /**
+     * OPERATOR ONLY: Updates the settings of an exchange. Reverts if exchange is not already addedNote: Need to pass in existing parameters if only
```
richardliang, line 564: Nit: space between added and Note:

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -469,6 +614,58 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
+        for (uint256 i = 0; i < _exchangeNames.length; i++) {
+
+            ExchangeSettings memory settings = exchangeSettings[_exchangeNames[i]];
+            uint256 maxTradeSize = isRipcord ? settings.incentivizedTwapMaxTradeSize : settings.twapMaxTradeSize;
+            uint256 slippageTolerance = isRipcord ? incentive.incentivizedSlippageTolerance : execution.slippageTolerance;
+        
+            bool isLever = newLeverageRatio > currentLeverageRatio;
+            LeverageInfo memory leverageInfo = _getAndValidateLeveragedInfo(slippageTolerance, maxTradeSize, _exchangeNames[i]);
+
+            (uint256 collateralNotional,) = _calculateChunkRebalanceNotional(leverageInfo, newLeverageRatio, isLever);
```
richardliang, line 656: Nit: typically we leave a space between punctuation `, )`

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -51,12 +52,26 @@ import { IChainlinkAggregatorV3 } from "../interfaces/IChainlinkAggregatorV3.sol
...
  * - Update _calculateBorrowUnits and _calculateMinRepayUnits to use chainlink as an oracle in 
+ *
+ * CHANGELOG 6/23/2021: 4eff1c0bcb721055b357cd6095eda3ed5cd6dd24
+ * - Add ExchangeSettings struct that contains exchange specific information
+ * - Update ExecutionSettings struct to not include exchange information
+ * - Add mapping of exchange names to ExchangeSettings structs and a list of enabled exchange names
+ * - Update constructor to take an array of exchange names and an array of ExchangeSettings
+ * - Add _exchangeName parameter to rebalancing functions to select which exchange to use
+ * - Add permissioned addEnabledExchange, updateEnabledExchange, and removeEnabledExchange functions
+ * - Add getTotalRebalanceNotional function
```
richardliang, line 63: Nit: getChunkRebalanceNotional

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -615,21 +615,25 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
-     * Calculates the total notional rebalance size. This can be used by external contracts and keeper bots to calculate the optimal exchange to rebalance with.
+     * Calculates the chunk rebalance size. This can be used by external contracts and keeper bots to calculate the optimal exchange to rebalance with.
      * Note: this function does not take into account timestamps, so it may return a nonzero value even when shouldRebalance would return ShouldRebalance.NONE for 
      * all exchanges (since minimum delays have not elapsed)
      *
-     * @return uint256      Total notional rebalance size. Measured in the asset that would be sold
-     * @return bool         Whether it would be levering or delevering
+     * @param _exchangeNames    Array of exchange names to get rebalance sizes for
+     *
+     * @return sizes            Array of total notional rebalance size. Measured in the asset that would be sold
```
bweick, line 627: `Array of total notional chunk size`

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -638,14 +642,28 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
+            LeverageInfo memory leverageInfo = _getAndValidateLeveragedInfo(slippageTolerance, maxTradeSize, _exchangeNames[i]);
 
-        bool isLever = newLeverageRatio > currentLeverageRatio;
-        uint256 leverageRatioDifference = isLever ? newLeverageRatio.sub(currentLeverageRatio) : currentLeverageRatio.sub(newLeverageRatio);
+            (uint256 collateralNotional,) = _calculateChunkRebalanceNotional(leverageInfo, newLeverageRatio, isLever);
 
-        uint256 collateralRebalanceNotional = leverageRatioDifference.preciseDiv(currentLeverageRatio).preciseMul(actionInfo.collateralBalance);
+            if (isLever) {
+                ActionInfo memory actionInfo = _createActionInfo();
+                sizes[i] = _calculateBorrowUnits(collateralNotional, actionInfo);
```
bweick: This should be `_calculateMaxBorrowCollateral`? I think we want to always be passing along the same asset

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
bweick: Think the amounts should always be the same asset, the collateral asset.

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
bweick: Actuall I don't think we need this logic. What's the thought behind adding this if statement? `_calculateChunkRebalanceNotional` already calculates the max borrow amount

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -533,6 +533,8 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
         _validateExchangeSettings(_exchangeSettings);
         
         exchangeSettings[_exchangeName] = _exchangeSettings;
+        exchangeSettings[_exchangeName].exchangeLastTradeTimestamp = 0;
```
bweick, line 536: Can we save gas by setting each variable individually instead of setting all of them and then over-writing? Not 100% sure this saves gas but assume it does

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
bweick: Open to reasons that it shouldn't be though

ncitron: We have it as the amount of the asset being sold. Since we are performing exact input trades, this is the relevant metric for keeper bots to calculate the optimal exchange.

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -645,16 +656,16 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
...
+
         for (uint256 i = 0; i < _exchangeNames.length; i++) {
 
-            ExchangeSettings memory settings = exchangeSettings[_exchangeNames[i]];
-            uint256 maxTradeSize = isRipcord ? settings.incentivizedTwapMaxTradeSize : settings.twapMaxTradeSize;
+            uint256 maxTradeSize = isRipcord ? 
+                exchangeSettings[_exchangeNames[i]].incentivizedTwapMaxTradeSize : 
+                exchangeSettings[_exchangeNames[i]].twapMaxTradeSize;
+                
             uint256 slippageTolerance = isRipcord ? incentive.incentivizedSlippageTolerance : execution.slippageTolerance;
```
bweick, line 670: Since this isn't used anywhere else let's just move this logic into the `LeverageInfo` declaration

bweick: I guess same goes for `maxTradeSize`

contracts/adapters/FlexibleLeverageStrategyAdapter.sol
```diff
@@ -670,7 +681,8 @@ contract FlexibleLeverageStrategyAdapter is BaseAdapter {
             (uint256 collateralNotional, ) = _calculateChunkRebalanceNotional(leverageInfo, newLeverageRatio, isLever);
 
             sizes[i] = isLever ? _calculateBorrowUnits(collateralNotional, leverageInfo.action) : collateralNotional;
```
bweick, line 684: Let's leave a note here that even though it says it's calculating units this function can also convert notional value as well.

contracts/adapters/FlexibleLeverageStrategyExtension.sol
```diff
@@ -36,7 +36,7 @@ import { StringArrayUtils } from "../lib/StringArrayUtils.sol";
 
 
 /**
- * @title FlexibleLeverageStrategyAdapter
+ * @title FlexibleLeverageStrategyExtension
  * @author Set Protocol
  *
  * Smart contract that enables trustless leverage tokens using the flexible leverage methodology. This adapter is paired with the CompoundLeverageModule from Set
```
bweick, line 42: Nit: there's a couple instances of using `adapter` in the javadoc. Let's just update those to `extension`

bweick (approved):

PR #65, opened 2021-07-06
Add FLIRebalanceViewer

ncitron:


contracts/viewers/FLIRebalanceViewer.sol
```diff
@@ -0,0 +1,253 @@
...
+     */
+    function shouldRebalanceWithBounds(
+        uint256 _customMinLeverageRatio,
+        uint256 _customMaxLeverageRatio
+    )
+        external
+        returns(string[] memory, FlexibleLeverageStrategyExtension.ShouldRebalance[] memory)
+    {
+
+        string[] memory enabledExchanges = fliStrategyExtension.getEnabledExchanges();
```
richardliang, line 101: Feel like we can dedupe some code here? We dont need to fetch enabled exchanges since line 225 (shouldRebalanceWithBounds) also returns the list of exchanges. We could set enabledExchanges to that first array and bring it up to this level

richardliang: Potentially one other way to organize the logic is to call FlexibleLeverageStrategyAdapter.shouldRebalanceWithBounds() first, then take the first return array (exchangeNames) and pass that into getChunkRebalanceNotional(), and then we get the `indexOf` the exchangeNames. This way we have all the data we need upfront to fetch quote and return into a priority order

richardliang (commented):
Overall logic looks good. Think we want to clean up / deduping more of the code. One pattern we like that you've seen already, is having a unified getter that fetches any data we need at the beginning and pass that to other internal functions (in a struct if needed).

contracts/viewers/FLIRebalanceViewer.sol
```diff
@@ -0,0 +1,258 @@
...
+            _maxLeverage
+        );
+
+        (actionInfo.uniV3Index, ) = actionInfo.exchangeNames.indexOf(uniswapV3ExchangeName);
+        (actionInfo.uniV2Index, ) = actionInfo.exchangeNames.indexOf(uniswapV2ExchangeName);
+
+        actionInfo.minLeverage = _minLeverage;
+        actionInfo.maxLeverage = _maxLeverage;
+
+        string[] memory exchangeNames = new string[](2);
```
richardliang, line 251: Just noticed this, we can delete this variable and use `actionInfo.exchangeNames` directly right

ncitron: We could use that, but then we'd have to refactor to use uniV3Index and uniV2Index when accessing chunkSendQuantiy. Both ways will work though. Should I make the change?

richardliang (approved):

PR #77, opened 2021-08-26
Fix Solidity warnings

cgewecke:
Small maintenance PR to make the solc compiler output more legible. Only changes comments, whitespace.

Removes dozens of warnings for SPDX (and handful of minor other things). There are still 4 license warnings coming from the Uniswap contracts we import from node_modules in the ExchangeIssuance contracts.

(Used Apache 2.0 for any files missing license info)


ncitron (approved):
Everything here looks good to me. Good to finally get rid of that giant block of errors on every compile.

PR #81, opened 2021-10-13
Intrinsic Productivity Rebalancing

ncitron:
STIP : [IP Rebalances](https://github.com/SetProtocol/ITIPS/pull/9)


contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+    /* ======== External Functions ======== */
+
+    /**
+     * Original function for starting rebalances from the inherited GIMExtension. This function has been deprecated
+     * in favor of startIPRebalance. It will revert when it is called.
+     */
+    function startRebalanceWithUnits(
+        address[] calldata /* _components */,
+        uint256[] calldata /* _targetUnitsUnderlying */,
+        uint256 /* _posotionMultiplier */
```
bweick, line 109: Nit: `positionMultiplier`

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+     * @param _transformComponent       component that needs to be transformed/untransformed during rebalance
+     * @param _transformInfo            TransfomInfo entry for the transform component
+     */
+    function setTransformInfo(address _transformComponent, TransformInfo memory _transformInfo) external onlyOperator {
+        require(
+            transformComponentInfo[_transformComponent].underlyingComponent == address(0),
+            "TransformInfo already set"
+        );
+        
+        require(_transformInfo.maxTransformSize <= type(uint96).max, "max transform size must be less than MAX_UINT_96");
```
bweick, line 132: Why use `uint96` here? Would add a comment explaining since that's a break from our usual typing.

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+    using PreciseUnitMath for uint256;
+    using SafeCast for int256;
+    using SafeCast for uint256;
+    using SafeMath for uint256;
+
+    /* ============ Structs =========== */
+
+    struct TransformInfo {
+        address underlyingComponent;        // underlying component address
+        uint256 maxTransformSize;           // max transform/untransform size measured in underlying component units
```
bweick, line 57: Why is there a max transform size? Seems like that should be just a wrap and unwrap action.

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+        uint256 minTransformDelay;          // minumum delay between transfroms/untransforms
```
bweick, line 58: Ditto why have a transform delay?

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+        ITransformHelper transformHelper;   // TransformHelper contract address
+    }
+
+    /* ========== State Variables ========= */
+
+    IAirdropModule public airdropModule;
+
+    // mapping from transform component to TransformInfo
+    mapping(address => TransformInfo) public transformComponentInfo;
+    // mapping from set component to target units
```
bweick, line 68: I'd maybe also include a note that each `TransformInfo` struct maps a transform component to its "set component". I do wonder if there's a better name than "set component", think there could be confusion what exactly is different between a Set component and a transform component

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+    mapping(address => uint256) public rebalanceParams;
```
bweick, line 69: Think this name can be more specific. It just maps to a unit number so I would just say that instead of general "params"

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+     *
+     * @param _setComponents            array of components involved in rebalance including components being removed (target units set to 0)
+     * @param _targetUnitsUnderlying    array of target units at end of rebalance, maps to same index of _components array
+     */
+    function startIPRebalance(address[] memory _setComponents, uint256[] memory _targetUnitsUnderlying) external onlyOperator {
+        require(_setComponents.length == _targetUnitsUnderlying.length, "length mismatch");
+        tradesComplete = false;
+
+        // clear out any current member of rebalanceParams from last rebalance
+        for (uint256 i = 0; i < setComponentList.length; i++) {
```
bweick, line 165: Is this necessary? If a component was removed in the previous rebalance (thus not part of this rebalance) it would already be 0. So you basically end up just removing and then adding back target units which seems inefficient.

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+     * @param _untransformData          array of untransform data 
+     */
+    function batchUntransform(
+        address[] memory _transformComponents,
+        bytes[] memory _untransformData
+    )
+        external
+        onlyAllowedCaller(msg.sender)
+    {
+        require(_transformComponents.length == _untransformData.length, "length mismatch");
```
bweick, line 191: Is there a way we can try to enforce `This function must be called after starting the rebalance, but before calling startTrades`?

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+
+        setComponentList = _setComponents;
+    }
+
+    /**
+     * ONLY ALLOWED CALLER: Untransforms components. This function must be called after starting the rebalance, but before calling startTrades.
+     * The untransformData parameter can be fetched by the rebalance bots by calling getUntransformData on the relevent TransformHelper.
+     *
+     * @param _transformComponents      array of components to untransform
+     * @param _untransformData          array of untransform data
```
bweick, line 182: What's included in this data?

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+    function _isTransformComponent(address _component) internal view returns (bool) {
+        return transformComponentInfo[_component].underlyingComponent != address(0);
+    }
+
+    /**
+     * Gets the total amount of a underlying component in the target set units. This value includes all units present due to
+     * a transform component containing it as its underlying. Units from the raw underlying component potentially being a component
+     * in the set are ignored.
+     */
+    function _getFinalTotalUnderlyingUnits(address _underlying, address[] memory _components) internal view returns (uint256) {
```
bweick, line 450: So this exists because we may wrap the same underlying component into multiple different assets?

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+
+            if (_isTransformComponent(component)) {
+                rebalanceTargets[i] = setToken.getDefaultPositionRealUnit(component).toUint256();
+            } else {
+
+                uint256 finalTotalUnderlyingUnits = _getFinalTotalUnderlyingUnits(component, setComponentList);
+                uint256 currentTotalUnderlyingUnits = _getCurrentTotalUnderlyingUnits(component, setComponentList);
+
+                uint256 targetUnderlying = rebalanceParams[component];
+                int256 diff = finalTotalUnderlyingUnits.toInt256() - currentTotalUnderlyingUnits.toInt256() + targetUnderlying.toInt256();
```
bweick, line 385: `SafeMath`?

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+            _transformComponent,
+            unitsToUntransform,
+            _untransformData
+        );
+
+        invokeManager(module, callData);
+    }
+
+    /**
+     * Untransforms a component
```
bweick, line 309: What is this function actually supposed to do? What is the math that should be happening? This javadoc tells me nothing about how this function is supposed to work

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+            _transformData
+        );
+
+        invokeManager(module, callData);
+    }
+
+    /**
+     * Parameterizes the rebalancing trades through GeneralIndexModule. For all transform components, the target units
+     * remain fixed at their current allocations. For non-transform units, the target units are calculated by taking into
+     * account the total amount of the component needed to create the target trasnform units.
```
bweick, line 367: What are all the components that make up the target non-transform units? There's some equation here I should be able to check against

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,478 @@
...
+                rebalanceTargets[i] = setToken.getDefaultPositionRealUnit(component).toUint256();
+            } else {
+
+                uint256 finalTotalUnderlyingUnits = _getFinalTotalUnderlyingUnits(component, setComponentList);
+                uint256 currentTotalUnderlyingUnits = _getCurrentTotalUnderlyingUnits(component, setComponentList);
+
+                uint256 targetUnderlying = rebalanceParams[component];
+                int256 diff = finalTotalUnderlyingUnits.toInt256() - currentTotalUnderlyingUnits.toInt256() + targetUnderlying.toInt256();
+
+                rebalanceTargets[i] = diff > 0 ? diff.toUint256() : 0;
```
bweick, line 387: What exactly is the `diff` param? Is this the amount of underlying tokens we expect to be left over after all the transforms are done at the end of the rebalance?

contracts/adapters/IPRebalanceExtension.sol
ncitron: Yep. This is already a requirement for some products in the pipeline, but does unfortunately complicate the accounting a bit

contracts/adapters/IPRebalanceExtension.sol
ncitron: This is because some transformations may have price impact. For example, a transformation could be a single sided deposit into curve using the amm module. To do this safely we need to be able to enforce a max sizes and delays much like the GIM

contracts/adapters/IPRebalanceExtension.sol
ncitron: See https://github.com/SetProtocol/index-coop-smart-contracts/pull/81#discussion_r771792435

contracts/adapters/IPRebalanceExtension.sol
ncitron: It is really just an intermediate value. The correct rebalance target is equal to max(diff, 0). I've added a lot more detail to the javadoc to help understand what is going on here.

contracts/adapters/IPRebalanceExtension.sol
ncitron: This is essentially an optional field that gets passed on to the TransformHelper which is free to use to covey any required additional information. One usecase I see for this is to allow the bots to supply minOut/maxIn amounts for transformations that can experience slippage such as amm deposits.

While I considered explicitly using a minOut and maxIn parameter, I chose this since it allow me to use EMPTY_BYTES as the value in cases where there is no slippage. I also figure that the world of yield in DeFi is quite complicated so it will be useful to have this as a bit of a wildcard.

What do you think? Would it be better to replace with maxIn / minOut and just ignore the value in the (most of the time) when slippage cannot happen?

contracts/adapters/IPRebalanceExtension.sol
ncitron: While it is not explicitly enforced, assuming all the untransformations were called during the untransform step, it will actually be impossible to call untransform again at a later point in the rebalancing process. This is because this call would be guaranteed to always correctly calculate that the amount to transform is 0, which causes it to revert with "nothing to untransform".

I could add an enum that represents the current stage of the rebalance process we are in and keep updating it as we go along. This would allow us to then directly enforce that things are being done at the correct step.

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,555 @@
...
+     * @param _transformComponent       component that needs to be transformed/untransformed during rebalance
+     * @param _transformInfo            new TransfomInfo entry for the transform component
+     */
+    function updateTransformInfo(address _transformComponent, TransformInfo memory _transformInfo) external onlyOperator {
+        require(
+            transformComponentInfo[_transformComponent].underlyingComponent != address(0),
+            "TransformInfo not set yet"
+        );
+
+        // Maximum transform size is limited to MAX_UINT_96 to prevent overflows when multiplying this maximum size by
```
Alpha-Guy, line 187: We have the same piece of code in this and the above function. We can add it in an internal function instead.

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,555 @@
...
+     * @param _untransformData          array of untransform data 
+     */
+    function batchUntransform(
+        address[] memory _transformComponents,
+        bytes[] memory _untransformData
+    )
+        external
+        onlyAllowedCaller(msg.sender)
+    {
+        require(_transformComponents.length == _untransformData.length, "length mismatch");
```
Alpha-Guy, line 227: Would suggest adding a "NONE" element in ENUM and checking it explicitly here, to verify we are in the right state.

Alpha-Guy: In general, would suggest doing the following in every function,
1. Verify we are in the right initial state, using the RebalanceStage function
2. Check other requirements and perform the actions corresponding to the function
3. Update the state to the next possible state

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,555 @@
...
+            _untransform(_transformComponents[i], _untransformData[i]);
+        }
+    }
+
+    /**
+     * ONLY OPERATOR: Begins the trading portion of the rebalance by initializing the GeneralIndexModule. Must be called after all required
+     * components have been untransformed.
+     */
+    function startTrades() external onlyOperator {
+        stage = RebalanceStage.TRADE;
```
Alpha-Guy, line 241: Let's set the TRADE state after successful completion of untransform functions, and verify we are in the TRADE state before executing startTrades.

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,555 @@
...
+        uint256 targetUnitsInTransformed = targetUnderlying.preciseMul(exchangeRate);
+        uint256 unitsToUntransform = currentUnits > targetUnitsInTransformed ? currentUnits.sub(targetUnitsInTransformed) : 0;
+        
+        uint256 maxUntransformUnits = transformInfo.maxTransformSize.preciseDiv(setToken.totalSupply()).preciseMul(exchangeRate);
+        if (unitsToUntransform > maxUntransformUnits) {
+            unitsToUntransform = maxUntransformUnits;
+        }
+
+
+        require(unitsToUntransform > 0, "nothing to untransform");
```
Alpha-Guy, line 342: We should not revert here. We should just return back on line `333` if `currentUnits <  targetUnitsInTransformed`, which is the case when `unitrsToUntransform` would be zero. Also, the reason we should not revert is it might break the logic in `batchUntransform`.

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,555 @@
...
+     * be untransformed. If this calculation yields a resulting amount to untransform that is larger than the
+     * maximum untransform size, the untransform size will be replaced with this max value. The rest of the
+     * transform component can be untransformed in later transactions.
+     */
+    function _untransform(address _transformComponent, bytes memory _untransformData) internal {
+        require(stage == RebalanceStage.UNTRANSFORM, "must be in untransform stage");
+
+        TransformInfo memory transformInfo = transformComponentInfo[_transformComponent];
+
+        require(transformInfo.minTransformDelay + lastTransform[_transformComponent] < block.timestamp, "delay not elapsed");
```
Alpha-Guy, line 319: All these requires can be grouped together and moved to an internal function, `validateUntransform`

contracts/adapters/IPRebalanceExtension.sol
```diff
@@ -0,0 +1,555 @@
...
+     * to transform is greater than maximum transform amount, the transform size will be replaced with the max value. The rest
+     * can be transformed in later transactions.
+     * 
+     */
+    function _transform(address _transformComponent, bytes memory _transformData, bool _transformRemaining) internal {
+        require(stage == RebalanceStage.TRANSFORM, "must be in transform stage");
+
+        TransformInfo memory transformInfo = transformComponentInfo[_transformComponent];
+
+        require(transformInfo.minTransformDelay + lastTransform[_transformComponent] < block.timestamp, "delay not elapsed");
```
Alpha-Guy, line 376: All the requires can be grouped together and moved to an internal function `validateTransform`

Alpha-Guy (changes requested):
The terminology used is a bit confusing in my opinion. Terms like `wrapped` and `underlying`, or `derivative` and `underlying` would have been better to differentiate between the different asset states. We could have used `sourceDerivative`, `underlying` and `targetDerivative` as well. 
While functions that transform the derivative to an underlying could have been called `derivativeToUnderlying()` or just `unwrap()` would have been enough, and functions that transform underlying to derivative could have been called `underlyingToDerivative()`, or just `wrap`. Again that's just my opinion.
The tests look good to me.

contracts/adapters/IPRebalanceExtension.sol
ncitron: There is unfortunately some problems with verifying that a stage has been completed on-chain. Since some transform/untransform steps can have slippage, it may be impossible to hit the target units to transform and untransform. This makes it difficult to check if the stage has been completed. We could add a percentage factor, so we could instead mark a stage as complete if we are within x% of the target. However, this would come at a significant gas cost, since we would now need to fetch all target units and position units for every single transform/untransform call. This will add a significant number of `SLOAD` and `CALL` operations, which tend to be pricey. We would run into the same problem when trying to check if the GeneralIndexModule rebalance is complete.

contracts/adapters/IPRebalanceExtension.sol
ncitron: See https://github.com/SetProtocol/index-coop-smart-contracts/pull/81#discussion_r773476469

contracts/adapters/IPRebalanceExtension.sol
ncitron: Also there is no `trade` function in this extension since we just use the `GeneralIndexModule` to do the trades. We could add a trade function to this extension that just checks the stage and forwards the call to `GeneralIndexModule`, but it still would not prevent an allowed caller for GIM from bypassing the check.

PR #85, opened 2021-11-24
Exchange issuance zero ex

ckoopmann:
Exchange Issuance of Set Tokens using 0x Exchange Aggregator.


ckoopmann:
Temporarily setting to draft while we are adjusting it to remove the intermediary WETH stage.

edkim:
Additional context from @ckoopmann: 
"There are some scripts that I used for development purposes, which could probably be removed. (A deploy script and the analysis of gas costs for the swaps). But I left it in for now to let the reviewers choose what should be removed."

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,582 @@
...
+    event ExchangeRedeem(
+        address indexed _recipient,     // The recipient address which redeemed the SetTokens
+        ISetToken indexed _setToken,    // The redeemed SetToken
+        IERC20 indexed _outputToken,    // The address of output asset(ERC20/ETH) received by the recipient
+        uint256 _amountSetRedeemed,     // The amount of SetTokens redeemed for output tokens
+        uint256 _amountOutputToken      // The amount of output tokens received by the recipient
+    );
+
+    event Refund(
+        address indexed _recipient,     // The recipient address which redeemed the SetTokens
```
ncitron, line 84: `Refund` is only emitted during an issuance, but the comment mentions redemption.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,582 @@
...
+        uint256 _refundAmount           // The amount of ETH redunded to the recipient
+    );
+
+    event BoughtTokens(IERC20 sellToken, IERC20 buyToken, uint256 boughtAmount);
```
ncitron, line 88: Emitting events costs gas, so we only want to do it if we thing its important data for analytics. I think we are safe to remove this, especially since all of this information can be derived by grabbing the amount from the `ExchangeIssue` event and the components from a call to the set token.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,582 @@
...
+         require(setController.isSet(address(_setToken)), "ExchangeIssuance: INVALID SET");
+         _;
+    }
+
+    modifier isWhitelistedIssuanceModule(address _issuanceModule) {
+        require(allowedIssuanceModules[_issuanceModule].isAllowed, "ExchangeIssuance: INVALID ISSUANCE MODULE");
+         _;
+    }
+
+    modifier isValidInput(ISetToken _setToken, uint256 _amountSetToken, ZeroExSwapQuote[] memory _componentQuotes) {
```
ncitron, line 102: What happens if a user passes in malformed data here? Can they use it to attack the contract? Will they lose their funds? If we pass in zero for the amount, it will revert on the issuance call (since both issuance modules already check if the amount is 0). If we pass in bad swaps, it will either fail when it tries to issue (and it hasn't bought the right components), or it will fail when we do the slippage check for redemption (since we likely would have forgotten to sell a component).

I think we should remove this. We can ensure all of these things at the frontend.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,582 @@
...
+    event Refund(
+        address indexed _recipient,     // The recipient address which redeemed the SetTokens
+        uint256 _refundAmount           // The amount of ETH redunded to the recipient
+    );
+
+    event BoughtTokens(IERC20 sellToken, IERC20 buyToken, uint256 boughtAmount);
+
+    /* ============ Modifiers ============ */
+
+    modifier isSetToken(ISetToken _setToken) {
```
ncitron, line 92: I think we don't need this either. Every issuance and redemption will check if  it is a valid set token anyway so it is redundant The only time where it can get weird is in `approveSetToken`, but I don't see any malicious activity that could be done by passing in a fake set token.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,582 @@
...
+        require(totalEthSold<= msg.value, "ExchangeIssuance: OVERSPENT ETH");
+        IBasicIssuanceModule(_issuanceModule).issue(_setToken, _amountSetToken, msg.sender);
+
+        uint256 amountEthReturn = msg.value.sub(totalEthSold);
+        if (amountEthReturn > 0) {
+            IWETH(WETH).withdraw(amountEthReturn);
+            (payable(msg.sender)).sendValue(amountEthReturn);
+        }
+
+        emit Refund(msg.sender, amountEthReturn);
```
ncitron, line 289: Actually I don't thing this refund event really makes sense to keep around anymore. In the old exchange issuance contract we added it since the refund was in ETH even when the input token was something else. Now this data feels pretty irrelevant since its just the difference between the maxInput (eth sent), and the eth used.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,582 @@
...
+    * Issues an exact amount of SetTokens for given amount of input ERC20 tokens.
+    * The excess amount of tokens is returned in an equivalent amount of ether.
+    *
+    * @param _setToken              Address of the SetToken to be issued
+    * @param _inputToken            Address of the input token
+    * @param _amountSetToken        Amount of SetTokens to issue
+    * @param _maxAmountInputToken   Amount of SetTokens to issue
+    * @param _componentQuotes       The encoded 0x transactions to execute 
+    *
+    * @return amountEthReturn       Amount of ether returned to the caller
```
ncitron, line 216: Shouldn't this be amountTokenReturn?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,582 @@
...
+        external
+        nonReentrant
+        returns (uint256)
+    {
+        _redeemExactSet(_setToken, _amountSetToken, _issuanceModule);
+        uint ethAmount = _sellComponentsForOutputToken(_setToken, _amountSetToken, _componentQuotes, IERC20(WETH), _issuanceModule);
+        require(ethAmount >= _minEthReceive, "ExchangeIssuance: INSUFFICIENT WETH RECEIVED");
+
+        IWETH(WETH).withdraw(ethAmount);
+        (payable(msg.sender)).sendValue(ethAmount);
```
ncitron, line 367: nit: do you need these extra parentheses? Won't `payable(msg.sender).sendValue(ethAmount)` work?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,582 @@
...
+contract ExchangeIssuanceZeroEx is Ownable, ReentrancyGuard {
+
+    using Address for address payable;
+    using SafeMath for uint256;
+    using PreciseUnitMath for uint256;
+    using SafeERC20 for IERC20;
+    using SafeERC20 for ISetToken;
+
+    struct ZeroExSwapQuote {
+        IERC20 sellToken;
```
ncitron, line 42: I don't think we need to include `sellToken` and `buyToken` and can just pass around `swapCallData` instead. While the structure is nice, this adds a ton of calldata to our calls which is relatively expensive gas wise. Being able to do the validations is nice, but even if the frontend supplies bad swaps, the call will likely just revert later in the transaction.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,582 @@
...
+        // Taken from: https://ethereum.stackexchange.com/a/111187/73805
+        if (!success) {
+            if (returndata.length == 0) revert();
+            assembly {
+                revert(add(32, returndata), mload(returndata))
+            }
+        }
+
+        boughtAmount = _quote.buyToken.balanceOf(address(this)).sub(buyTokenBalanceBefore);
+        spentAmount = sellTokenBalanceBefore.sub(_quote.sellToken.balanceOf(address(this)));
```
ncitron, line 513: Can we calculate the spend amount at the very end of the for loops that use `_fillQuote`? Since we mostly only care about the total amount spent, we can just compare the contracts balance before/after all the quotes have been filled. This will save us a ton of gas since `balanceOf` is an external call which is pretty expensive.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,582 @@
...
+        emit BoughtTokens(_quote.sellToken, _quote.buyToken, boughtAmount);
+    }
+
+    /**
+     * Redeems a given amount of SetToken.
```
ncitron, line 518: Can we add a note to this comment to explain that it also transfers the token from the caller?

ncitron (commented):
Everything looks really good here. Added some comments mostly focused on gas efficiency.

scripts/calculateEIGasCosts.ts
```diff
@@ -0,0 +1,127 @@
+import { ethers } from "hardhat";
```
ckoopmann, line 1: This is a script we used in development to compare the efficiency of :
1. Trading all of the input token into WETH and then from WETH into each of the components (using `1 + numComponents` number of 0x api quotes)
2. Trading directly between input token and the components. Letting 0x figure out the route to go for each trade. (Using `numComponent` number of api quotes). This is what we ended up implementing.

Please let me know if I should remove / move this script

scripts/deployExchangeIssuanceZeroEx.ts
```diff
@@ -0,0 +1,28 @@
+import { ethers } from "hardhat";
```
ckoopmann, line 1: This is an ad hoc deploy script that I used to deploy the contracts to my local mainnet fork when I started to work on the frontend integration for index-ui (which is still WIP).

Let me know if I should move / remove this.

test/exchangeIssuance/exchangeIssuance.spec.ts
```diff
@@ -499,7 +498,9 @@ describe("ExchangeIssuance", async () => {
       context("when input erc20 token is weth", async () => {
         cacheBeforeEach(async () => {
           subjectInputToken = weth;
-          await subjectInputToken.connect(subjectCaller.wallet).approve(exchangeIssuance.address, MAX_UINT_256, { gasPrice: 0 });
+          await subjectInputToken
```
ckoopmann, line 501: FYI: I had to change all of these tests because I updated hardhat (to be able to fork the latest mainnet block, see integration tests below), which meant that we couldn't use "gasPrice: 0" anymore. So wherever we set `gasPrice: 0` in the tests to check ether balance changes accurately, I instead used my newly defined helper function to get the transaction fee and account for that in the tests.

test/integration/exchangeIssuanceZeroExIntegration.spec.ts
```diff
@@ -0,0 +1,442 @@
+import "module-alias/register";
```
ckoopmann, line 1: This is an integration test script that is meant to be run manually against a local mainnet fork.

Since it uses the response from the real 0x api it needs to be executed against a fork that is as "up to date" with the latest block as possible. 

To run these tests you should:
1. Start local mainnet fork with `FORK=true LATESTBLOCK=true yarn chain`
2. Run integrationtest with `INTEGRATIONTEST=true VERBOSE=true yarn test test/integration/exchangeIssuanceZeroExIntegration.spec.ts` (leave out verbosity flag if you don't want to see details on the 0x api requests / response)

test/integration/exchangeIssuanceZeroExIntegration.spec.ts
```diff
@@ -0,0 +1,442 @@
...
+  isDebtIssuanceModule: boolean;
+};
+
+type TokenName = "SimpleToken" | "DPI";
+
+function logVerbose(...args: any[]) {
+  if (process.env.VERBOSE) console.log(...args);
+}
+
+if (process.env.INTEGRATIONTEST) {
```
ckoopmann, line 36: I added this flag to make extra sure that these tests are not run in the normal test / ci pipeline. Not sure if actually necessary

ckoopmann (commented):
Some context for the reviewers.

test/integration/exchangeIssuanceZeroExIntegration.spec.ts
ckoopmann: This test can also be used as inspiration for building a frontend integration.

test/integration/exchangeIssuanceZeroExIntegration.spec.ts
ckoopmann: The reasons why I didn't want them to be run in the normal ci:
1. The local chain they are run against is different (forking from the latest block)
2. The tests are not necessarily reproducible since they depend on the 0x api response, which is outside of our control.

Point 2.  could be changed by generating / saving a bunch of 0x api quotes for a given block and then run the tests against that fixed block number with those quotes. 

However the point of these integration tests for me was to simulate a workflow as close as possible of how it would be used in production, and manually play with the parameters.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Completely removed this event since it is probably not necessary with the same logic as the BoughtTokens event. (Also we quite inconsistently did not have a refund event for ERC20 refunds without even noticing it).

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Removed it

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: True, I don't think it is security relevant. Might save people some gas by reverting earlier. But since these conditions are not dependant on contract state they should come up during "gasEstimation" already anyway, before a transaction is submitted.

Will remove it as suggested

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Agreed. We do call state changing methods on this address (`approve`) but I guess there is no reentrancy risk or anything from this.

Removed the modifier accordingly

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Correct. Actually I renamed it to `totalInputTokenSold` to be consistent with the variable name returned in the function body

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Correct, removed those parantheses

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Done. Also added a note that the components that the set is redeemed for will be received on this contracts address.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Actually we do need to check the balances of each component token separately to avoid the user using "leftover" component amounts in the contract for their issuance / redemption. However for the input/output token one balance call at the end of the loop should indeed be enough.

I will move these balance checks outside of the fillQuote method so that we can reduce the number accordingly (from `2*numComponents` to `numComponents + 1`).

ckoopmann: I moved the checks of the balance changes outside the fillQuote function and changed it so that the difference of the Input/Output Token balance is only calculated once.

Fell free to check the changes in this commit to see if this is what you had in mind:
https://github.com/IndexCoop/index-coop-smart-contracts/pull/85/commits/af03e21cd9757e9a7fc9d8ca0ca31c1e31206300

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: As suggested I removed the Struct and switched to passing the callData around directly.

Feel free to check this commit to see if this is what you had in mind:
https://github.com/IndexCoop/index-coop-smart-contracts/pull/85/commits/b5d9bc134ff43eddaff6943cba31becfe37bcd10

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+    using SafeERC20 for ISetToken;
+
+    struct IssuanceModuleData {
+        bool isAllowed;
+        bool isDebtIssuanceModule;
+    }
+
+    /* ============ Constants ============== */
+
+    address constant public ETH_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
```
cgewecke, line 48: It would be nice to have comment about what this is, e.g that it's a NULL address placeholder used to identify ETH as if it was an ERC20 token (for example when emitted as an event arg or when passed in as a parameter to toggle ETH specific logic).

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+        for (uint256 i = 0; i < _issuanceModuleAddresses.length; i++) {
+            _addIssuanceModule(_issuanceModuleAddresses[i], _issuanceModuleDebtIssuanceFlags[i]);
+        }
+
+    }
+
+    /* ============ External Functions ============ */
+
+    /**
+     * Withdraw slippage to selected address
```
cgewecke, line 108: + What is the use case for this method? 
+ How does the contract get into a state where tokens / ETH need to be swept from it?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+                _to.sendValue(address(this).balance);
+            }
+            else{
+                _tokens[i].safeTransfer(_to, _tokens[i].balanceOf(address(this)));
+            }
+        }
+    }
+
+    receive() external payable {
+        // required for weth.withdraw() to work properly
```
cgewecke, line 125: How does `receive` get called as part of weth.withdraw? Is this actually about weth.deposit failing?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+     * @param _setToken    Address of the SetToken being initialized
+     */
+    function approveSetToken(ISetToken _setToken, address _issuanceModule) external {
+        address[] memory components = _setToken.getComponents();
+        for (uint256 i = 0; i < components.length; i++) {
+            // Check that the component does not have external positions
+            require(
+                _setToken.getExternalPositionModules(components[i]).length == 0,
+                "ExchangeIssuance: EXTERNAL_POSITIONS_NOT_ALLOWED"
+            );
```
cgewecke, line 194: Could this require be moved to `approveToken`? Otherwise you can approve a component and bypass the external positions check.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+
+    /**
+     * Runs all the necessary approval functions required for a given ERC20 token.
+     * This function can be called when a new token is added to a SetToken during a
+     * rebalance.
+     *
+     * @param _token    Address of the token which needs approval
+     */
+    function approveToken(IERC20 _token, address _spender) public  isWhitelistedIssuanceModule(_spender) {
+        _safeApprove(_token, _spender, type(uint96).max);
```
cgewecke, line 166: I see this `uint96` max value in the V1 version of the contract as well. Why is this being used (as opposed to MAX_UINT256?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+     * Sets a max approval limit for an ERC20 token, provided the current allowance
+     * is less than the required allownce.
+     *
+     * @param _token    Token to approve
+     * @param _spender  Spender address to approve
+     */
+    function _safeApprove(IERC20 _token, address _spender, uint256 _requiredAllowance) internal {
+        uint256 allowance = _token.allowance(address(this), _spender);
+        if (allowance < _requiredAllowance) {
+            _token.safeIncreaseAllowance(_spender, type(uint96).max - allowance);
```
cgewecke, line 387: Is accidental overflow possible here? `allowance` and `_requiredAllowance` are `uint256` but the max is `96` ?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+    }
+
+    /**
+    * Issues an exact amount of SetTokens for given amount of input ERC20 tokens.
+    * The excess amount of tokens is returned in an equivalent amount of ether.
+    *
+    * @param _setToken              Address of the SetToken to be issued
+    * @param _inputToken            Address of the input token
+    * @param _amountSetToken        Amount of SetTokens to issue
+    * @param _maxAmountInputToken   Amount of SetTokens to issue
```
cgewecke, line 206: Param description is mistaken here. Original contract comment was:

> Maximum amount of input tokens to be used to issue SetTokens.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+        bytes[] memory _componentQuotes,
+        address _issuanceModule
+    )
+        isWhitelistedIssuanceModule(_issuanceModule)
+        external
+        nonReentrant
+        payable
+        returns (uint256)
+    {
+        require(msg.value > 0, "ExchangeIssuance: INVALID ETH AMOUNT");
```
cgewecke, line 262: Nit: Could this be clarified to `...: NO ETH SENT` since that's what is checked here?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+
+        IWETH(WETH).deposit{value: msg.value}();
+        _safeApprove(IERC20(WETH), swapTarget, msg.value);
+
+        uint256 totalEthSold = _buyComponentsForInputToken(_setToken, _amountSetToken, _componentQuotes, IERC20(WETH), _issuanceModule);
+
+        require(totalEthSold<= msg.value, "ExchangeIssuance: OVERSPENT ETH");
```
cgewecke, line 269: Nit: add space before `<=`

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+        isWhitelistedIssuanceModule(_issuanceModule)
+        external
+        nonReentrant
+        returns (uint256)
+    {
+
+        uint256 outputAmount;
+        _redeemExactSet(_setToken, _amountSetToken, _issuanceModule);
+
+        // Liquidate components for WETH and ignore _outputQuote
```
cgewecke, line 311: What does "ignore _outputQuote" refer to here? Is this still true?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+    function _safeApprove(IERC20 _token, address _spender, uint256 _requiredAllowance) internal {
+        uint256 allowance = _token.allowance(address(this), _spender);
+        if (allowance < _requiredAllowance) {
+            _token.safeIncreaseAllowance(_spender, type(uint96).max - allowance);
+        }
+    }
+
+    /**
+     * Issues an exact amount of SetTokens using WETH.
+     * Acquires SetToken components at the best price accross uniswap and sushiswap.
```
cgewecke, line 393: Using 0x rather...

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+            else {
+                uint256 componentBalanceBefore = IERC20(component).balanceOf(address(this));
+                _fillQuote(_quotes[i]);
+                uint256 componentBalanceAfter = IERC20(component).balanceOf(address(this));
+                componentAmountBought = componentBalanceAfter.sub(componentBalanceBefore);
+                require(componentAmountBought >= units, "ExchangeIssuance: UNDERBOUGHT COMPONENT");
+            }
+        }
+        uint256 inputTokenBalanceAfter = _inputToken.balanceOf(address(this));
+        totalInputTokenSold = totalInputTokenSold.add(inputTokenBalanceBefore.sub(inputTokenBalanceAfter));
```
cgewecke, line 425: Not certain about this but might it be cheaper to return the value directly rather than declare and set `totalInputTokenSold`?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+    }
+    
+    /**
+     * Whitelists an issuance module
+     *
+     * @param _issuanceModule    Struct containing data on issuance module to add
+     */
+    function _addIssuanceModule(address _issuanceModule, bool _isDebtIssuanceModule) internal {
+        allowedIssuanceModules[_issuanceModule].isAllowed = true;
+        allowedIssuanceModules[_issuanceModule].isDebtIssuanceModule = _isDebtIssuanceModule;
```
cgewecke, line 365: Out of curiosity, which features of DebtIssuanceModule would people be using here? Am asking just because DI is typically used for tokens with external positions (which this contract prohibits).

cgewecke (commented):
Hi @ckoopmann :)

Am just visiting from SetProtocol with a short review. This looks good and seems like a nice simplification of the original contracts as well. 

Have left a couple questions (mostly about things ported over from EI-V1). 

(Am going to look through the tests for this this afternoon, just scanned them and they look cool)

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+
+        (address[] memory components, uint256[] memory componentUnits) = getRequiredIssuanceComponents(_issuanceModule, _setToken, _amountSetToken);
+
+        uint256 inputTokenBalanceBefore = _inputToken.balanceOf(address(this));
+        for (uint256 i = 0; i < components.length; i++) {
+            address component = components[i];
+            uint256 units = componentUnits[i];
+
+            // If the component is equal to the input token we don't have to trade
+            if(component == address(_inputToken)) {
```
cgewecke, line 412: The logic inside this `if` block is not getting hit by the unit tests.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+        (address[] memory components, uint256[] memory componentUnits) = getRequiredRedemptionComponents(_issuanceModule, _setToken, _amountSetToken);
+        uint256 outputTokenBalanceBefore = _outputToken.balanceOf(address(this));
+        for (uint256 i = 0; i < _swaps.length; i++) {
+            uint256 maxAmountSell = componentUnits[i];
+
+            uint256 outputTokenAmountBought;
+            uint256 componentAmountSold;
+
+            // If the component is equal to the input token we don't have to trade
+            if(components[i] == address(_outputToken)) {
```
cgewecke, line 451: The logic inside this `if` block is not getting hit by the unit tests.

test/integration/exchangeIssuanceZeroExIntegration.spec.ts
```diff
@@ -0,0 +1,420 @@
...
+    let wbtcUnits: BigNumber;
+
+    const setTokenScenarios: Partial<Record<TokenName, SetTokenScenario>> = {};
+    // Test Parameterization
+    const SET_TOKEN_NAMES: TokenName[] = ["DPI", "SimpleToken"];
+    const SET_TOKEN_AMOUNTS: Record<TokenName, number[]> = {
+      SimpleToken: [1],
+      // DPI issuance of 5000 failed due to underbuying Badger in a multihop
+      // TODO: Investigate
+      DPI: [3000, 1],
```
cgewecke, line 66: Good test!

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+        _removeIssuanceModule(_issuanceModuleAddress);
+    }
+
+    /**
+     * Change the _swapTarget
+     *
+     * @param _swapTarget    Address of the swap target contract. (Usually ZeroEx ExchangeProxy)
+     */
+    function setSwapTarget(address _swapTarget) public onlyOwner {
+        swapTarget = _swapTarget;
```
cgewecke, line 155: 0x has been very consistent in the addressing of their exchange proxy. It's "0xdef1c0ded..." for every network (except 1337)

https://github.com/0xProject/protocol/blob/development/packages/contract-addresses/addresses.json

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+
+        (bool success, bytes memory returndata) = swapTarget.call(_quote);
+
+        // Forwarding errors including new custom errors
+        // Taken from: https://ethereum.stackexchange.com/a/111187/73805
+        if (!success) {
+            if (returndata.length == 0) revert();
+            assembly {
+                revert(add(32, returndata), mload(returndata))
+            }
```
cgewecke, line 490: Works! (Saw a `UniswapFeature/UnderBought` error while running the integration tests)

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+     * Acquires SetToken components at the best price accross uniswap and sushiswap.
+     * Uses the acquired components to issue the SetTokens.
+     *
+     * @param _setToken          Address of the SetToken being issued
+     * @param _amountSetToken    Amount of SetTokens to be issued
+     *
+     */
+    function _buyComponentsForInputToken(ISetToken _setToken, uint256 _amountSetToken, bytes[] memory _quotes, IERC20 _inputToken, address _issuanceModule) internal returns (uint256 totalInputTokenSold) {
+        uint256 componentAmountBought;
+        uint256 inputTokenAmountSold;
```
cgewecke, line 402: Declared but unused.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+    function _sellComponentsForOutputToken(ISetToken _setToken, uint256 _amountSetToken, bytes[] memory _swaps, IERC20 _outputToken, address _issuanceModule)
+        internal
+        returns (uint256 totalOutputTokenBought)
+    {
+        (address[] memory components, uint256[] memory componentUnits) = getRequiredRedemptionComponents(_issuanceModule, _setToken, _amountSetToken);
+        uint256 outputTokenBalanceBefore = _outputToken.balanceOf(address(this));
+        for (uint256 i = 0; i < _swaps.length; i++) {
+            uint256 maxAmountSell = componentUnits[i];
+
+            uint256 outputTokenAmountBought;
```
cgewecke, line 447: Declared but unused.

scripts/calculateEIGasCosts.ts
cgewecke: What were the discoveries of this investigation?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Good point, added the respective comment

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: We call `weth.withdraw` when we want to return ETH to the user.

This is done since for ETH trades (i.e. when using ETH as input / output taken during issue/redeem) we actually send a WETH based trade to the 0x Proxy / Router. To return ETH instead of WETH to the user in those cases we call `withdraw` which in turn sends eth to this contract triggering the `receive`.

ckoopmann: Withdrawing eth to return leftover from issuance:
https://github.com/FlattestWhite/index-coop-smart-contracts/blob/exchange-issuance-zero-ex/contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol#L274

Withdrawing eth to send out resulting funds from redemption:
https://github.com/FlattestWhite/index-coop-smart-contracts/blob/exchange-issuance-zero-ex/contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol#L350

ckoopmann: @cgewecke Does that answer your question ?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Good question. tbh I kind of "blindly" copied this, but I vaguely remember wondering about this as well.
@ncitron Do you have any insight on this ?
If not I will just change it to max uint256 I guess.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Uups, thanks. Adjusted back to the original comment

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Absolutely. Changed accordingly

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,561 @@
...
+    /**
+     * Issues an exact amount of SetTokens using WETH.
+     * Acquires SetToken components at the best price accross uniswap and sushiswap.
+     * Uses the acquired components to issue the SetTokens.
+     *
+     * @param _setToken          Address of the SetToken being issued
+     * @param _amountSetToken    Amount of SetTokens to be issued
+     *
+     */
+    function _buyComponentsForInputToken(ISetToken _setToken, uint256 _amountSetToken, bytes[] memory _quotes, IERC20 _inputToken, address _issuanceModule) internal returns (uint256 totalInputTokenSold) {
```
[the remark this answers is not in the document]
ckoopmann: Done

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Good question. 
@ncitron requested adding support for different Issuance modules (debt and basic) so maybe he can elaborate on the exact use case ?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Glad to hear that. Will give the SO where I took this from an upvote for this. 😅
I previously tried adding a test contract that returns a solidity `CustomError` for testing this in the unit tests but gave up on it after a day since the necessary hardhat / ethers upgrades to support solidity 0.8.4 got me into all kinds of trouble.

test/integration/exchangeIssuanceZeroExIntegration.spec.ts
ckoopmann: Thanks, just noticed that the TODO comment above this line is outdated as well.

ckoopmann: Removed the todo comment

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Good to know, didn't know this. (Nor do I understand how they did this 😅).

Is this just FYI or do you suggest any changes to our implementation based on this ?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: This is outdated and refered to a  previous implementation when we swapped the components to WETH and then WETH to input/output token in separate 0x trades.

ckoopmann: Removed it

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Adjusted. Also the param list was incomplete which i fixed.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: This variable is used to be able to account for the case when one of the components is also the input / output token. (see `if(component == address(_inputToken))...` above. In this case the change in input token balance alone would not account for the total const of the issuance. 

Hope that makes sense. If not let me know and I will rephrase.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Thanks, I guess I should be paying more attention to solidity compile warnings😅.

Removed it.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Thanks, removed it.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: See:
https://github.com/FlattestWhite/index-coop-smart-contracts/blob/exchange-issuance-zero-ex/contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol#L413

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: For ERC20 tokens it can happen if the 0x swaps somehow return a higher amount of component tokens than is required to issue the set. Also of course someone could mistakenly send ERC20 tokens to this contract address.

For ETH actually I don't see any obvious way right now for it to get stuck in the contract.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Like so much else, I copied this from ExchangeIssuanceV1. imo it could be moved to approveToken.
@ncitron Do you see any reason why not to move this to `approveToken` ?

ckoopmann: Also I would like to double check that the `approveToken` methods are supposed to be available to be called by anyone. (I assume so, since it this is also copy-paste from EI V1)

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Good point. Again this is copy paste from EI V1. Actually my `git blame` returns @cgewecke as the author of the equivalent lines in those contracts. So not sure who might be able to clear this up.

I guess the answer to this depends in part on the outcome of the discussion of why we are using MAX_UINT96 in the first place.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Good catch. I added a test case where the input token is one of the components to cover this.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Good catch. I added a test case where the output token is one of the components to cover this.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
cgewecke: Yes, thanks so much for this explanation :100:

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
cgewecke: IIRC I am not the actual author (not sure how I got commits on this - perhaps via V2 which I copied V1 for). Noah definitely knows this contract the best of any of us I think.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
cgewecke: Ah yes sorry! lgtm.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ncitron: Originally it was done since some tokens (like COMP) have a maximum amount of MAX_UINT_96. However, I never liked this since even in the COMP implementation, if you pass in MAX_UINT_256, it would still work. I'm ok with changing it.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ncitron: Right now GMI uses DIM since it is removable, unlike BasicIssuanceModule. We want to be able to support issuing GMI

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ncitron: I was under the impression that 0x upgraded their swap router from time to time? I thought this would be a nice addition since it would allow us to not have to redeploy whenever 0x had a contract upgrade. If immutability is a big concern I would be open to removing this.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ncitron: How could we know that it is an external position in `approveToken`? It does not take a set token as a parameter

ckoopmann: Good point. @cgewecke I guess that answers the question, and I will just leave this as is ?

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: I just went ahead and changed it to `type(uint256).max`.|

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: As mentioned above I replaced `type(uint96).max` with `type(uint256).max` so overflow should not be an issue.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: @cgewecke I assume this answers your question. If not feel free to reopen the thread.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: The router is a "Diamond" - Proxy so I don't think they regularly deploy that, since they will just redeploy / upgrade the implementation contracts for different feature. The current instance of the router / proxy was deployed June 2020. 

However I agree it might be wasteful to redeploy the whole Exchange Issuance if / when that proxy address changes in the future. 

Also I'd think the risk is manageable since we don't really hold user funds in the EI contract,  only approve the 0x Router with the amount necessary for each trade and use token balances to verify traded amounts. (As opposed to relying on the return value of the swap call).

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
cgewecke: Ok! Makes sense.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ncitron: Actually since its a proxy (and unlikely to change), I think we should make it immutable. Allowing it to be changed to an arbitrary address allows the owner to steal funds from any user who has approved the exchange issuance contract as a spender. This can be done by setting an ERC20 address as the router, then doing an exchange issuance with an encoded transferFrom call to rip tokens out of past users addresses. In fact, an attack like this has happened to both DYDX and Gelato recently.

cgewecke: Oh!! Wow good eye @ncitron.

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
cgewecke: (@ckoopmann Would it be ok to un-resolve this? Am fine with leaving things as they but would just like to flesh out all the implications and make sure I understand this....)

ncitron: So `approveToken` generically approves any token to be spent by a particular issuance module. The purpose of this is to save gas when new components are added to a set token. However, there is nothing special we can do to gatekeep what kind of tokens get approved. Even if we wanted to add a `setToken` parameter to `approveToken`, someone could just initialize a brand new set with any token they want in it as a default position (even if it is an external position for another set), and have it be approved.

Also, I'm not sure that we need to disallow tokens that are external positions. Since we now have DIM support, they should just work out of the box? I actually think it might be best to remove the check entirely. What do you think @cgewecke?

cgewecke: > Since we now have DIM support, they should just work out of the box? 

@ncitron @ckoopmann Ok that sounds right to me. Agree with removal.

ckoopmann: I removed the external position check 👍

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: Great point. I made the swapTarget immutable

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
```diff
@@ -0,0 +1,519 @@
...
+
+    /**
+     * Returns component positions required for issuance 
+     *
+     * @param _issuanceModule    Address of issuance Module to use 
+     * @param _setToken          Set token to issue
+     * @param _amountSetToken    Amount of set token to issue
+     */
+    function getRequiredIssuanceComponents(address _issuanceModule, ISetToken _setToken, uint256 _amountSetToken) public view returns(address[] memory components, uint256[] memory positions) {
+        try IDebtIssuanceModule(_issuanceModule).getRequiredComponentIssuanceUnits(_setToken, _amountSetToken)
```
ckoopmann, line 484: As suggested I removed the issuance module whitelist. 
Since we previously used the issuance module registration to keep track of wether it is a Basic or Debt-Issuancemodule, I switched to just using this try / catch block, which seems to work but might not be optimal. (For example it will "swallow" an revert reason that might be thrown in  `getRequiredComponentIssuanceUnits` and it is probably also not gas efficient).

@ncitron Let me know if you have an idea for a better way of determining wether a given issuance module is basic or debt issuance. (Of course we can also bring back a mapping in this contract but I thought the point was to avoid that).

cgewecke (commented):
This all looks good to me (although there's still a remaining question about module handling in the getters). 

Nice work!

contracts/exchangeIssuance/ExchangeIssuanceZeroEx.sol
ckoopmann: As @ncitron suggested I removed the try/catch logic and added an `_isDebtIssuance` flag to the redeem / issuance methods to let the user specify wether given issuance module is basic or debt issuance.

ncitron (approved):
LGTM! Feel free to merge @ckoopmann

PR #89, opened 2022-01-17
WIP: Leveraged Token exchange issuance

ckoopmann:


contracts/exchangeIssuance/ExchangeIssuanceLeveraged.sol
```diff
@@ -0,0 +1,621 @@
...
+        returns (uint256)
+    {
+        (address longToken, uint256 longAmount,,) = getLeveragedTokenData(_setToken, _amountSetToken);
+
+        address[] memory assets = new address[](1);
+        assets[0] = IAToken(longToken).UNDERLYING_ASSET_ADDRESS();
+        uint[] memory amounts =  new uint[](1);
+        amounts[0] = longAmount;
+
+        bytes memory params = abi.encode(_setToken, _amountSetToken, msg.sender, true, _paymentToken, _paymentParams);
```
ncitron, line 316: Why is `_paymentParams` encoded as bytes then encoded again here? I think we can just leave it as a uint

contracts/exchangeIssuance/ExchangeIssuanceLeveraged.sol
```diff
@@ -0,0 +1,621 @@
...
+
+
+    /**
+     * Swaps the debt tokens obtained from issuance for the underlying of the collateral
+     *
+     * @param _setToken                   Address of the SetToken to be issued
+     * @param _setAmount                  Amount of SetTokens to issue
+     * @param _longTokenUnderlying        Address of the underlying of the collateral token
+     */
+    function _swapShortForLongTokenUnderlying(address _setToken, uint256 _setAmount, address _longTokenUnderlying) internal returns (uint256) {
```
ncitron, line 384: Can we instead allow the user to select what exchange to use using an enum? This removes the gas costs of needing to figure out the best exchange on-chain, and lets us add new exchanges much more easily.

contracts/exchangeIssuance/ExchangeIssuanceLeveraged.sol
```diff
@@ -0,0 +1,621 @@
...
+        IERC20(_longTokenUnderlying).approve(address(LENDING_POOL), _depositAmount);
+        LENDING_POOL.deposit(_longTokenUnderlying, _depositAmount, address(this), 0);
+    }
+
+
+    /**
+     * Approves lending pool to transfer owed amount of borrowed tokens after flashloan operation is complete
+     *
+     */
+    function _approveAssetToReturn(
```
ncitron, line 411: We can probably do this in `approveSetToken` (and we might want to remove `approveToken` since its probably not important to be able to approve one by one). This prevents us from doing a new approval on each issuance which saves a but of gas (not that it matters much on optimism)

ncitron (commented):
This is a good start. Left some minor comments but everything looks like its going in the right direction.

ckoopmann:
Will close this PR to avoid confusion and reopen a new PR later when the whole Leveraged EI feature is ready for review by Set Team

PR #91, opened 2022-02-02
Autopublish releases to npm

FlattestWhite:


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@setprotocol/index-coop-contracts",
+  "name": "@indexcoop/index-coop-contracts",
```
0xModene, line 2: should be 
```typescript
"name": "@indexcoop/index-coop-smart-contracts",
```

0xModene (changes requested):
small change needed but mostly looks good

package.json
FlattestWhite: whoops - fixed

FlattestWhite:
Just a note on this PR - circle ci atm isn't actually verifying this repo because the yarn.lock wasn't regenerated when porting it over. We cache based on the yarn.lock file. Since the yarn.lock file remains the same, the old build and test steps still pass but its pointing to ~/index-coop-private

0xModene:
@FlattestWhite Can you run `yarn` locally and resolve this?

FlattestWhite:
> @FlattestWhite Can you run `yarn` locally and resolve this?

done

0xModene (approved):

PR #92, opened 2022-02-02
Add chainlink keepers interface for rebalancing

FlattestWhite:
First implementation of FliRebalanceKeeper.

For the FlexibleLeverageExtension, we need to remove the nonEOA and probably add nonRetrancy guard or something.


contracts/keepers/FliRebalanceKeeper.sol
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * performUpkeep checks that a rebalance is required. Otherwise the contract call will revert.
+     */
+    function performUpkeep(bytes calldata /* performData */) external override onlyRegistry {
+        Address.functionCall(fliExtension, getRebalanceCalldata());
+    }
+
+    function getRebalanceCalldata() private returns (bytes memory) {
+        bytes memory shouldRebalanceCalldata = abi.encodeWithSignature("shouldRebalance()");
```
ncitron, line 65: Can we create a new interface called for the leverage extension so we don't need to manually encode with strings here?

contracts/keepers/FliRebalanceKeeper.sol
```diff
@@ -0,0 +1,79 @@
...
+        bytes memory shouldRebalanceResponse = Address.functionCall(address(fliExtension), shouldRebalanceCalldata, "Failed to execute shouldRebalance()");
+        (string[] memory exchangeNames, uint256[] memory shouldRebalances) = abi.decode(shouldRebalanceResponse, (string[], uint256[]));
+
+        uint256 shouldRebalance = shouldRebalances[0];
```
ncitron, line 69: We might want to be able to select what exchange to use. Maybe we can store what index to use somewhere?

contracts/keepers/FliRebalanceKeeper.sol
```diff
@@ -0,0 +1,79 @@
...
+        if (shouldRebalance == 1) {
+            return abi.encodeWithSignature("rebalance(string)", exchangeNames[0]);
```
ncitron, line 71: If we had created an interface like in my previous comments, we could replace this with

```solidity
abi.encodeWithSelector(fliExtension.rebalance.selector, exchangeNames[0])
```

Which is a bit cleaner

contracts/keepers/FliRebalanceKeeper.sol
```diff
@@ -0,0 +1,79 @@
...
+    constructor(address _fliExtension, address _registryAddress) public {
+        fliExtension = _fliExtension;
+        registryAddress = _registryAddress;
+    }    
+
+    /**
+     * As checkUpkeep is not a view function, calling this function will actually consume gas.
+     * As such if a keeper calls this function, it will always return true so that performUpkeep will be called.
+     */    
+    function checkUpkeep(bytes calldata /* checkData */) external override onlyRegistry returns (bool, bytes memory) {
```
ncitron, line 53: Even though checkUpkeep isn't a view function, it is called as if it were by the keepers. This is because you can call a non-view method and it runs as if it were executed, except state changes are not reflected. We can then just have this function do the same thing as `getRebalanceCalldata`. This will save us on gas since it can be run off-chain.

Finally I don't think we need to make this onlyRegistry since it won't perform any gatekeeped actions.

contracts/keepers/FliRebalanceKeeper.sol
```diff
@@ -0,0 +1,79 @@
+/*
+    Copyright 2021 Index Cooperative.
```
ncitron, line 2: 2022

ncitron (commented):
Good start! Left a few comments.

contracts/keepers/FliRebalanceKeeper.sol
FlattestWhite: Shouldn't the fliExtension contract determine which exchange to use? Rather than this contract? Or are you suggesting to have a whitelist of exchanges that this contract can call?

contracts/keepers/FliRebalanceKeeper.sol
FlattestWhite: done

contracts/keepers/FliRebalanceKeeper.sol
FlattestWhite: done

contracts/keepers/FliRebalanceKeeper.sol
FlattestWhite: done - removed it

contracts/keepers/FliRebalanceKeeper.sol
FlattestWhite: done

contracts/keepers/FliRebalanceKeeper.sol
ncitron: At the moment we allow the caller of the FLI extension to determine what exchange to use (of a list of whitelisted exchanges). This is because determining the optimal exchange may be difficult or gas extensive to do on-chain. Lately we have just been hardcoding our bots to use the exchange we want, with a backup exchange already whitelisted in case we need to make changes quickly. For this contract since we have a bit less trust in the chainlink keepers, we probably want to hard code which exchange to use.

FlattestWhite: Hmmm - then we keep a list of allowable addresses that we can contact? What's the exchangeName that comes back from the fliExtension call - I'm assuming it's an address?

ncitron: The exchange name is actually just a string which corresponds to the adapter that is used to do the trade part of the rebalance. I think its probably best to hardcode the index of the array that we want to use (and allow the owner of the contract to change it).

FlattestWhite:
ready for re-review @ncitron added the exchange index parameter

ncitron (commented):
Looking pretty good. I just noticed one major change related to how we want to pass data between `checkUpkeep` and `performUpkeep`. Also we need to add support for allowing smart contracts to make rebalance calls on `AaveLeverageStrategyExtension`.

Everything else is looking pretty good. Great job on the tests.

contracts/keepers/FliRebalanceKeeper.sol
```diff
@@ -0,0 +1,90 @@
...
+        _;
+    }
+
+    /* ============ State Variables ============ */
+
+    IFlexibleLeverageStrategyExtension public fliExtension;         // Address of the fli extension contract
+    address public registryAddress;                                 // Address of the chainlink keeper registry
+    uint256 public exchangeIndex;                                   // The index of the exchange to use
+
+    /* ============ Constructor ============ */
```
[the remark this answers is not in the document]
FlattestWhite: done

contracts/keepers/FliRebalanceKeeper.sol
```diff
@@ -0,0 +1,90 @@
...
+     * performUpkeep checks that a rebalance is required. Otherwise the contract call will revert.
+     */
+    function performUpkeep(bytes calldata performData) external override onlyRegistry {
+        require(performData.length > 0, "Invalid performData");
+        Address.functionCall(address(fliExtension), performData);
+    }
+
+    function getRebalanceCalldata() private returns (bytes memory) {
+        bytes memory shouldRebalanceCalldata = abi.encodeWithSelector(fliExtension.shouldRebalance.selector);
+        bytes memory shouldRebalanceResponse = Address.functionCall(address(fliExtension), shouldRebalanceCalldata, "Failed to execute shouldRebalance()");
```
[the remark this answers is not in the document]
FlattestWhite: Added a LeverageSettings struct with options for bounds and added it as part of the constructor

contracts/keepers/FliRebalanceKeeper.sol
```diff
@@ -0,0 +1,90 @@
...
+        bytes memory callData = getRebalanceCalldata();
+        return (callData.length > 0, callData);
+    }
+
+    /**
+     * performUpkeep checks that a rebalance is required. Otherwise the contract call will revert.
+     */
+    function performUpkeep(bytes calldata performData) external override onlyRegistry {
+        require(performData.length > 0, "Invalid performData");
+        Address.functionCall(address(fliExtension), performData);
```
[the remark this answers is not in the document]
FlattestWhite: Done - made the checkUpkeep return an encoded `shouldRebalance` enum and exchangeName. This is then decoded in performUpkeep before sending a transaction off to the fliExtension

contracts/keepers/FliRebalanceKeeper.sol
```diff
@@ -0,0 +1,90 @@
...
+
+    IFlexibleLeverageStrategyExtension public fliExtension;         // Address of the fli extension contract
+    address public registryAddress;                                 // Address of the chainlink keeper registry
+    uint256 public exchangeIndex;                                   // The index of the exchange to use
+
+    /* ============ Constructor ============ */
+    constructor(IFlexibleLeverageStrategyExtension _fliExtension, address _registryAddress) public {
+        fliExtension = _fliExtension;
+        registryAddress = _registryAddress;
+        exchangeIndex = 0;
```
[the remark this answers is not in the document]
FlattestWhite: yep done

ncitron (commented):
This all looks great. We still need to update `AaveLeverageStrategyExtension` to allow non-EOA accounts to call it. I think the best way to do this is to update https://github.com/IndexCoop/index-coop-smart-contracts/blob/master/contracts/lib/BaseExtension.sol#L59 to check some stored flag that determines whether we want to skip the EOA check. Similar to how `onlyAllowCaller` works here https://github.com/IndexCoop/index-coop-smart-contracts/blob/master/contracts/lib/BaseExtension.sol#L145

FlattestWhite:
> This all looks great. We still need to update `AaveLeverageStrategyExtension` to allow non-EOA accounts to call it. I think the best way to do this is to update https://github.com/IndexCoop/index-coop-smart-contracts/blob/master/contracts/lib/BaseExtension.sol#L59 to check some stored flag that determines whether we want to skip the EOA check. Similar to how `onlyAllowCaller` works here https://github.com/IndexCoop/index-coop-smart-contracts/blob/master/contracts/lib/BaseExtension.sol#L145

Added a boolean to allow us to bypass the onlyEOA check. Although from the documentation, the check is there to stop sandwich/flash loan attacks. Wouldn't flipping the flag open us up to these type of attacks? Do we need to store an allowlist of contract addresses that are allowed to call this function?

ncitron (approved):

PR #100, opened 2022-03-16
Add Curve Support in Leveraged EI 

ckoopmann:


contracts/exchangeIssuance/ExchangeIssuanceLeveraged.sol
```diff
@@ -116,6 +114,9 @@ contract ExchangeIssuanceLeveraged is ReentrancyGuard, FlashLoanReceiverBaseV2,
 
     /* ============ Modifiers ============ */
 
+    // TODO: Review if we really need this check.
+    // In the 0x case we see a lot of people using approveTokens to use the contract with non-index-coop sets
+    // Maybe we should make their life easier and just remove this check? Are there any security implications?
```
ckoopmann, line 119: @ncitron : What do you think about this ? Can we just remove `isSetToken` ? Do you see any security implications ?

ncitron: Yeah I think this is fine to remove. Execution would fail anyway later down the line when it tries to issue/redeem with a non set token

contracts/exchangeIssuance/DEXAdapter.sol
```diff
@@ -21,147 +21,176 @@ import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
...
     using SafeMath for uint256;
 
     /* ============ Constants ============= */
 
     uint256 constant private MAX_UINT256 = type(uint256).max;
-    uint24 public constant POOL_FEE = 3000;
+    address public constant ETH_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+    // This is the margin in wei we add on top of stEth amounts to avoid errors. 
+    // This is needed both in aave as well as curve integration 
+    // TODO: Review to find better solution
```
ckoopmann, line 53: @ncitron : This is the rounding issue I mentioned in slack. Which I also encountered when doing a curve exact-output-swap (which lead to revertion due to the swap underbuying by 1 or 2 wei).

Unless you veto it, I'd like to just keep this as is now and remove the TODO comment.

ncitron: Yeah I agree that this is a fine concession to make. Nobody is going to miss 2 wei. I think the problem has something to do with astETH rebases

contracts/exchangeIssuance/DEXAdapter.sol
```diff
@@ -254,31 +284,261 @@ abstract contract DEXAdapter {
...
+
+        return ICurveCalculator(_addresses.curveCalculator).get_dx(
+            poolData.nCoins,
+            poolData.balances,
+            poolData.A,
+            poolData.fee,
+            poolData.rates,
+            poolData.decimals,
+            // TODO: Review correct value for "underlying" parameter
+            false,
```
ckoopmann, line 439: According to the parameter [documentation in the deployed contract ](https://etherscan.io/address/0xc1DB00a8E5Ef7bfa476395cdbcc98235477cDE4E#code#L158) this parameter indicates `Whether the coin is in raw or lent-out form`. 

For the stEth/Eth usecase it works this way and I don't see any use case where we would wan to swap a "lent-out" token via curve.

Therefore I'd like to keep this hardcoded. Alternatively we could add this as an additional field in the `swapData` struct to be controlled by the user, but I don't think that's necessary.

@ncitron : Let me know if you disagree

ncitron: Yeah I also agree on this

ncitron (approved):
LGTM! Can you just remove the TODO comments leftover from the thread before you merge

PR #101, opened 2022-03-23
feat: leverage token exchange issuance quote functions

ncitron:


contracts/exchangeIssuance/DEXAdapter.sol
```diff
@@ -188,6 +182,90 @@ library DEXAdapter {
...
+                _swapData.path[0],
+                _swapData.path[1],
+                ICurveAddressProvider(_addresses.curveAddressProvider)
+            );
+            return _getAmountOutCurve(_swapData.pool, i, j, _amountIn, _addresses);
+        } else {
+            return _getAmountOutUniV2(
+                _swapData,
+                IUniswapV2Router02(
+                    (_swapData.exchange == Exchange.Quickswap) ? _addresses.quickRouter : _addresses.sushiRouter
```
ckoopmann, line 220: Now that we use this check in so many places we might want to move it into a `pure` helper function, for better maintainability / readability.

contracts/exchangeIssuance/DEXAdapter.sol
```diff
@@ -419,7 +496,7 @@ library DEXAdapter {
         uint256 _amountOut,
         Addresses memory _addresses
     )
-        public
```
ckoopmann, line 422: Ups, forgot to turn those back to private after debugging.
Well spotted 👍

contracts/exchangeIssuance/ExchangeIssuanceLeveraged.sol
```diff
@@ -28,17 +28,14 @@ import { IAaveLeverageModule } from "../interfaces/IAaveLeverageModule.sol";
 import { IDebtIssuanceModule } from "../interfaces/IDebtIssuanceModule.sol";
 import { IController } from "../interfaces/IController.sol";
 import { ISetToken } from "../interfaces/ISetToken.sol";
+import { IQuoter } from "../interfaces/IQuoter.sol";
```
ckoopmann, line 31: I don't think you are using the `IQuoter` in this contract. Could probably remove the import.

contracts/exchangeIssuance/ExchangeIssuanceLeveraged.sol
```diff
@@ -217,6 +214,65 @@ contract ExchangeIssuanceLeveraged is ReentrancyGuard, FlashLoanReceiverBaseV2{
         _approveToken(_token);
     }
 
+    /**
+     * Gets the input cost of issuing a given amount of a set token. This
+     * function is not marked view, but should be static called from frontends.
+     * This constraint is due to the need to interact with the Uniswap V3 quoter
```
ckoopmann, line 220: Even if we were not interacting with UniV3 we still couldn't have this as view since we are syncing the aaveLeverage module. Up to you if  you want to adjust the comment.

contracts/exchangeIssuance/ExchangeIssuanceLeveraged.sol
```diff
@@ -217,6 +214,65 @@ contract ExchangeIssuanceLeveraged is ReentrancyGuard, FlashLoanReceiverBaseV2{
         _approveToken(_token);
     }
 
+    /**
```
ckoopmann, line 217: Maybe we should mention somewhere in the comments that ther will be an error if the swap paths overlap. (In terms of exchange and path), the size of which depends on the liquidity.

test/exchangeIssuance/exchangeIssuanceLeveraged.spec.ts
```diff
@@ -805,6 +825,201 @@ describe("ExchangeIssuanceLeveraged", async () => {
...
+            await performIssuance();
+            const finalAmount = await subjectInputToken.balanceOf(owner.address);
+
+            const expectedAmountIn = initAmount.sub(finalAmount);
+
+            expect(amountIn).to.gt(preciseMul(expectedAmountIn, ether(0.99)));
+            expect(amountIn).to.lt(preciseMul(expectedAmountIn, ether(1.01)));
+          });
+
+          context("when using Uniswap V3 is the exchange", async () => {
```
ckoopmann, line 893: Probably a typo copied from myself but I guess it should be "as the exchange".

test/integration/ethereum/exchangeIssuanceLeveragedIntegration.spec.ts
```diff
@@ -78,6 +78,7 @@ if (process.env.INTEGRATIONTEST) {
           addresses.dexes.uniV2.router,
           addresses.dexes.sushiswap.router,
           addresses.dexes.uniV3.router,
+          addresses.dexes.uniV3.quoter,
```
ckoopmann, line 81: I didn't see a Curve test case anywhere. If you want to add it, it would probably easiest to be done here since here. (In the unittests you'd have to spin up the curve setup)

ckoopmann (approved):
Mostly LGTM. Only smaller nits + request for a Curve testcase (assuming that I haven't overlooked an already existing one).

PR #106, opened 2022-06-27
Gmspacex/hardhat abi config

gmspacex:
In this PR, added `hardhat-abi-exporter` package and updated `hardhat.config.ts`
```
  abiExporter: {
    path: "./abi",
    runOnCompile: true,
    clear: true,
    flat: true,
    spacing: 2,
  },
```

### DIFF
--- a/abi/AaveLeverageStrategyExtension.json
+++ b/abi/AaveLeverageStrategyExtension.json
@@ -1,0 +1,1344 @@
+[
+  {
+    "inputs": [
+      {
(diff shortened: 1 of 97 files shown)


ckoopmann (commented):
As discussed, I'm not a fan of tracking build artifacts in git, as it will pollute the repo with a large number of file, while bringing little benefit.

Note that the abis are already generated at build time in the hardhat compile artifacts. 
Also we generate typechain classes which provide more typesafety etc. when interacting with contracts.

PR #109, opened 2022-07-05
Flash Mint Leveraged Compound Integration

muranox:


ckoopmann (changes requested):
Looks promising 👍 

Please add tests for these contracts that show that they are actually working as expected.

We need both unit tests running against our own mock-setup as well as integration tests against real deployments.

You can probably copy the leveraged integration tests and adjust them slightly to make it work for this contract.

Examples:
[Integration test](https://github.com/IndexCoop/index-coop-smart-contracts/blob/master/test/integration/ethereum/exchangeIssuanceLeveragedIntegration.spec.ts)

[UnitTests](https://github.com/IndexCoop/index-coop-smart-contracts/blob/master/test/exchangeIssuance/exchangeIssuanceLeveraged.spec.ts)

After the tests are green I'll give it another more detailed / thorough review.

muranox:
Just created another pull request

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+    * Sets various contract addresses
+    *
+    * @param _weth                  Address of wrapped native token
+    * @param _quickRouter           Address of quickswap router
+    * @param _sushiRouter           Address of sushiswap router
+    * @param _uniV3Router           Address of uniswap v3 router
+    * @param _uniV3Quoter           Address of uniswap v3 quoter
+    * @param _setController         SetToken controller used to verify a given token is a set
+    * @param _debtIssuanceModule    DebtIssuanceModule used to issue and redeem tokens
+    * @param _compoundLeverageModule    CompoundLeverageModule to sync before every issuance / redemption
```
ckoopmann, line 159: Nit:  columns of the param documentation not aligned.

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+        addresses.quickRouter = _quickRouter;
+        addresses.sushiRouter = _sushiRouter;
+        addresses.uniV3Router = _uniV3Router;
+        addresses.uniV3Quoter = _uniV3Quoter;
+        addresses.curveAddressProvider = _curveAddressProvider;
+        addresses.curveCalculator = _curveCalculator;
+
+        address _cEtherADDRESS = CompoundLeverageModuleStorage(address(_compoundLeverageModule)).underlyingToCToken(_weth);
+        require(_cEtherADDRESS != address(0x0), "CEtherAddress Error");
+        cEtherADDRESS = _cEtherADDRESS;
```
ckoopmann, line 194: Nit: change to `cEtherAddress` for consistency

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+     *
+     * @param _collateralToken              Address of collateral token
+     * @param _depositAmount                Amount to deposit
+     */
+    function _depositToCompound(
+        address _collateralToken,
+        uint256 _depositAmount
+    ) internal {
+        // get cTokenAddress from underlying assets
+        address cTokenAddress = _getCTokenAddess(_collateralToken);
```
ckoopmann, line 1200: The cToken address is already in the `LeveragedTokenData` of the `DecodedParams` so you could probably just pass it through here instead of regenerating it from the underlying. This would probably be cheaper in gas as well as reduce complexity / code size.

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+            (components, equityPositions, debtPositions) = debtIssuanceModule.getRequiredComponentRedemptionUnits(_setToken, _setAmount);
+        }
+
+        
+        require(equityPositions[0] == 0 || equityPositions[1] == 0, "ExchangeIssuance: TOO MANY EQUITY POSITIONS");
+        
+        require(debtPositions[0] == 0 || debtPositions[1] == 0, "ExchangeIssuance: TOO MANY DEBT POSITIONS");
+
+        if(equityPositions[0] > 0){
+            if (components[0] == cEtherADDRESS) {
```
ckoopmann, line 604: This (the whole outer if / else block) looks like a lot of code duplication to me. The difference in the two if/else branches seems only to be `components[0]` vs `components[1]` etc.  I think you could simplify this code a if you assigned the collateral  / debt token in the beginning. Might even save some gas since you'd be doing a lot fewer array accesses.

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+    )
+    internal
+    {
+        // need to check (true or false to issue)
+        compoundLeverageModule.sync(_setToken, true);
+
+        LeveragedTokenData memory leveragedTokenData = _getLeveragedTokenData(_setToken, _setAmount, true);
+
+        // get cTokenAddress from underlying assets
+        _getCTokenAddess(leveragedTokenData.collateralToken);
```
ckoopmann, line 669: You don't use the returned value here so I guess you can remove this statement.

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+     * @param _collateralAmount      Amount of collateralToken to withdraw
+     */
+    function _withdrawFromCompound(
+        address _collateralToken,
+        uint256 _collateralAmount
+    ) internal returns (uint256){
+        uint256 result;
+
+        // get cTokenAddress from underlying assets
+        address cTokenAddress = _getCTokenAddess(_collateralToken);
```
ckoopmann, line 1225: As mentioned elsewhere you could just pass that data through from the `LeveragedTokenData` instead, which is cheaper and easier.

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+        }
+        return result;
+    }
+
+    /**
+     * Returns the cToken Address from underlying asset Address
+     *
+     * @param _collateralToken       Address of the the collateral token
+     */
+    function _getCTokenAddess(
```
ckoopmann, line 1249: Once you have replaced the usages of this function with just using the existing data, you can remove this function, as well as the `CompoundLeverageModuleStorage` interface

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ckoopmann: I think it'd also be fine to just pass `_cEtherAddress` in as constructor argument, which would reduce complexity.

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+        // get cTokenAddress from underlying assets
+        address cTokenAddress = _getCTokenAddess(_collateralToken);
+
+        if (_collateralToken != addresses.weth) {
+            IERC20(_collateralToken).approve(cTokenAddress, _depositAmount);
+            ICErc20Delegator(cTokenAddress).mint(_depositAmount);
+        } else {
+            IWETH(addresses.weth).withdraw(_depositAmount);
+            ICEther(payable(cTokenAddress)).mint{value: _depositAmount}();
+            // return mintTokens;
```
ckoopmann, line 1208: Nit: Remove commented out code.

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+    internal
+    view
+    returns (LeveragedTokenData memory)
+    {
+        address[] memory components;
+        uint256[] memory equityPositions;
+        uint256[] memory debtPositions;
+        address _collateralToken;
+        LeveragedTokenData memory _leveragedTokenData;
+        uint256 requiredCTokenAmount;
```
ckoopmann, line 590: Maybe rename this variable just `cTokenAmount` since it isn't really "required" in the redemption case, rather it is the expected amount received upon redemption.

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+        address _collateralToken,
+        uint256 _collateralAmount
+    ) internal returns (uint256){
+        uint256 result;
+
+        // get cTokenAddress from underlying assets
+        address cTokenAddress = _getCTokenAddess(_collateralToken);
+        
+        Exp memory exchangeRate = Exp({mantissa: ICEther(payable(cTokenAddress)).exchangeRateStored()});
+        uint256 cTokenAmount = div_((_collateralAmount), exchangeRate);
```
ckoopmann, line 1228: I think here you are just reversing the calculation from cTokenAmount that you did in `_getLeveragedTokenData`
Instead it might be easier / more efficient to just attach the `cTokenAmount` as an additional field to the `LeveragedTokenData` and pass it down here.

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
```diff
@@ -0,0 +1,1346 @@
...
+
+        uint256 balance = ICErc20Delegator(cTokenAddress).balanceOf(address(this));
+        require(balance >= cTokenAmount, "ExchangeIssuance: Collateral Amount issue");
```
ckoopmann, line 1231: I don't think we need this check, since the withdrawal from compound would fail anyway. Even though it might be a slight help in debugging I don't think it is worth all that used gas. (external contract call + additional variable + require)

test/exchangeIssuance/exchangeIssuanceLeveragedForCompound.spec.ts
```diff
@@ -0,0 +1,1122 @@
+import "module-alias/register";
```
ckoopmann, line 1: As mentioned on discord we also need integration tests in addition to these unit tests.

contracts/exchangeIssuance/CErc20Storage.sol
```diff
@@ -0,0 +1,24 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    SPDX-License-Identifier: Apache License, Version 2.0
+*/
+pragma solidity 0.6.10;
+pragma experimental ABIEncoderV2;
+
+contract CErc20Storage {
```
ckoopmann, line 19: It seems you are using this contract only as interface. In that case I'd declare it as such and move it to the `interfaces` folder. (same for `CompoundLeverageModuleStorage` if you don't remove that one entirely - see comments below)

ckoopmann (changes requested):
Implementation looks pretty good overall, some minor nits as well as some suggestions regarding potential simplifications / gas savings.

The main remaining task would be adding the integration test suite.
After that I'll take another look over the testing as a whole.

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: Seems resolved by 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: > Nit: change to `cEtherAddress` for consistency

Seems resolved by 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: Seems resolved by 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: addressed in 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: addressed in 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: addressed in 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: addressed in 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: addressed in 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: addressed in 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: addressed in 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: addressed in 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/CErc20Storage.sol
ElmoArmy: addressed in 2bc43dfd123d9c826c22e1931048af595bb3ebc3

contracts/exchangeIssuance/ExchangeIssuanceLeveragedForCompound.sol
ElmoArmy: > I think it'd also be fine to just pass `_cEtherAddress` in as constructor argument, which would reduce complexity.

I agree with you on this change and it has not been implemented.

ckoopmann (approved):

ElmoArmy:
After discussing with @ckoopmann, it looks like we have three outstanding tasks on this PR.

- [ ] Rename instances of ExhangeIssuance to FlashMint in the name of files/contracts/functions. 
- [ ] Rename instances of ExhangeIssuance to FlashMint in comments/tests.
- [ ] Pass in _cEtherAddress in as constructor argument.

PR #116, opened 2022-08-31
Flash Mint contract for Set tokens with wrapped components

bergben:
Implements flash mint functionality for Set tokens with wrapped components. 
~~Wrapping info is passed in as calldata. Only allowlisted wrap-callTargets are enabled (and when wrap callTarget == set token component, e.g. for CompoundV2).~~ 
Uses wrap adapters directly, only the integrationName is passed in which is used to fetch the wrap adapter directly from the IntegrationRegistry.
This Flash mint contract is generalized and can be used for any protocol, given that WrapAdapters behave the same way.

This contract also supports flash minting for Set tokens that contain both unwrapped and wrapped components (technically speaking it would even support token sets with no wrapped components at all).
 Does not support debt positions on Set token.
 Wrapping / Unwrapping is skipped for a component if ComponentSwapData[component_index].underlyingERC20 address == set component address.

If the set contains both the wrapped and unwrapped version of a token (e.g. DAI and cDAI) -> two separate component data points have to be supplied in _swapData and _wrapData
 e.g. for issue
 Set components at index 0 = DAI; then -> ComponentSwapData[0].underlyingERC20 = DAI; (no wrapping will happen)
 Set components at index 1 = cDAI; then -> ComponentSwapData[1].underlyingERC20 = DAI; (wrapping will happen)

 **Compatible with:**
 IssuanceModules: DebtIssuanceModule, DebtIssuanceModuleV2, IWrapV2Adapter
 
**Caveat**
There might be cases where some wrapped components dust may stay in the flashMintContract because the exchange rate or swap rate might change a tiny bit. Setting a tiny tolerance is the task of the caller... 
There is still the check of maxAmountInputSpent / minAmountOutput
This is because the exchange rate from wrapped to unwrapped is used for computation of the actual amounts moments before the actual transaction with the input data is triggered.
This could only be solved by passing in a selector method that returns the exchange rate for each wrapped token (to directly compute everything just-in-time on-chain). It gets complicated... and costs more gas...
The contract is withdrawable, so that dust can be collected


.prettierignore
```diff
@@ -1 +0,0 @@
-contracts/
```
bergben, line 1: Is there a reason why solidity contracts are ignored for prettier? 

maybe it didn't work well enough back when the repo was created? 

Having a consistent formatting across the code base seems valuable

bergben:
@FlattestWhite could you give this a first look? Is it in line with the thoughts you had for the implementation of this?

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,905 @@
...
 
-    event ExchangeIssue(
-        address indexed _recipient,     // The recipient address of the issued SetTokens
-        ISetToken indexed _setToken,    // The issued SetToken
-        IERC20 indexed _inputToken,     // The address of the input asset(ERC20/ETH) used to issue the SetTokens
-        uint256 _amountInputToken,      // The amount of input tokens used for issuance
-        uint256 _amountSetIssued        // The amount of SetTokens received by the recipient
-    );
+  IController public immutable setController;
+  IDebtIssuanceModule public immutable debtIssuanceModule; // interface is compatible with DebtIssuanceModuleV2
```
bergben, line 88: I don't think this has to be backwards compatible for any previous set Tokens? I assumed we don't have to support the BasicIssuanceModule or WrapAdapters of V1 here

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,905 @@
...
 
-    /* ============ Constants ============= */
+  struct ComponentSwapData {
+    // unwrapped token version, e.g. DAI
+    address underlyingERC20;
+    // swap data for DEX operation: fees, path, etc. see DEXAdapter.SwapData
+    DEXAdapter.SwapData dexData;
+    // ONLY relevant for issue, not used for redeem:
+    // amount that has to be bought of the unwrapped token version (to cover required wrapped component amounts for issuance)
+    // this amount has to be computed beforehand through the exchange rate of wrapped Component <> unwrappedComponent
```
bergben, line 70: If WrapAdapters would include a getter for this exchange rate, we could do this in a more elegant way without requiring the buy amounts of the Underlying (unwrapped) token.... 

Compound, Aave, [...] all should support such functionality. 

We could do this on our own Set fork (i.e. WrapAdapterV3), but that would mean yet another change to original contract code right from the start

contracts/exchangeIssuance/FlashMintWrapped.sol
FlattestWhite: I used the debt issuance module here as it's upgradeable versus the basic issuance module. @ckoopmann for the zero ex one, that's iirc the reason why we use the debt issuance module there

.prettierignore
FlattestWhite: yeah it's a legacy thing we inherited from Set labs. Adding consistent formatting using prettier sgtm

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,905 @@
...
+   */
+  function _wrapComponent(
+    address _wrappedComponent,
+    uint256 _wrapAmount,
+    address _underlyingToken,
+    ComponentWrapData calldata _wrapData
+  ) internal {
+    // 1. get the wrap adapter directly from the integration registry
+
+    // Note we could get the address of the adapter directly in the params instead of the integration name
```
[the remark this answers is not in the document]
FlattestWhite: See ExchangeIssuanceZeroEx where we're forced to pass in the bytes calldata for the initial swap function. Truth to be told, the reason I want to directly avoid using the wrap adapters is to save a lot of gas on all the contract calls. In saying that, we could actually have a view function in this contract which uses the wrap adapters to generate all the swap data params and then have the client pass the raw bytes passed into the execute function.

However, we'll still need to take in raw bytes for the initial exchange (such as with 0x protocol).

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: yes that was also my thinking. it also has hooks and stuff :D basicIssuanceModule is a deadlock in current form, can't be removed

contracts/exchangeIssuance/FlashMintWrapped.sol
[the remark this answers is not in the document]
bergben: I think you are right about saving gas, I just ran some rough numbers and getting the adapter and call data costs about 15k gas.

The difference with ExchangeIssuanceZeroEx is that we _know_ and _define_ the swapTarget on which we invoke.
I am not really comfortable with calling out to an unknown contract for security reasons.

_But_ there is a simple solution to that: 
For Compound and Yearn the wrap / unwrap calls are invoked directly on the wrappedToken address. 
For Aave there is one lendingPool contract which all wrapping / unwrapping methods go to.
So:
`if wrapCallTarget == SetToken component ||  allowedWrapCallTargets[wrapCallTarget]`
should do the trick.

We add a mapping allowedWrapCallTargets that can be edited by the contract owner.
SetToken components are indirectly allowListed, someone would have to be able to add components to the SetToken

No drawbacks on security + gas cost saved... What do you think?

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: Edward mentioned that there is a BIMv2 basically implemented https://github.com/SetProtocol/set-protocol-v2/pull/178 

We might want to go with that one for CMI, to save gas cost etc. 

I'd also change the method names in BIMv2 to be in line with DIMv2

contracts/exchangeIssuance/FlashMintWrapped.sol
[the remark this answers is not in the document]
bergben: @FlattestWhite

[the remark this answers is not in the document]
FlattestWhite: Yeah having an allowlist of addresses that we're allowed to call was what I was thinking as well. An SLOAD is cheaper than a contract call after all. We'll have to make this contract pausable, ownable so that we can add and remove different allowlisted contracts.

contracts/exchangeIssuance/FlashMintWrapped.sol
FlattestWhite: sgtm

contracts/exchangeIssuance/FlashMintWrapped.sol
FlattestWhite: Yeah i'm thinking with all the logic around exchange rates and the prices on the initial swap to/from WETH, can we push that calculation to the frontend? such that we have the view method take in all the quantities etc and just spit out the relevant calldata that can be passed back into the mint/redeem function?

contracts/exchangeIssuance/FlashMintWrapped.sol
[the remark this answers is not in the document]
bergben: yea I have it Withdrawable (which includes Ownable) 

hmm why Pausable @FlattestWhite? No other Set / IndexCoop contract is Pausable AFAIK

bergben:
@FlattestWhite I changed the description of the PR. Please have a look at the caveat and let me know what you think...

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,1087 @@
...
+  ) internal {
+    require(_invokeWrapData.callTarget != address(0) && _invokeWrapData.callData[0] != 0,"FlashMint: WRAP_DATA_MISSING");
+
+    require(_invokeWrapData.callTarget == _wrappedToken || wrapTargetsAllowList[_invokeWrapData.callTarget],"FlashMint: WRAP_CALLTARGET_NOT_ALLOWED");
+
+    // approve _underlyingToken token transfer from this to _wrapCallTarget
+    DEXAdapter._safeApprove(IERC20(_underlyingToken), _invokeWrapData.callTarget, _wrapAmount);
+    // invoke wrap function call. we can't check any response value because the implementation might be different 
+    // between wrapCallTargets... e.g. compoundV2 would return uint256 with value 0 if successful
+    _invokeWrapData.callTarget.functionCallWithValue(_invokeWrapData.callData, _invokeWrapData.value);
```
ckoopmann, line 986: This looks dangerous to me. `_invokeWrapData.callData` is provided by the user right ? So couldn't he encode a transfer / approve call here to potentially drain leftover funds from other users ?

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,1087 @@
...
-     *
-     * @param _setToken          Address of the SetToken being initialized
-     * @param _issuanceModule    Address of the issuance module which will be approved to spend component tokens.
-     */
-    function approveSetToken(ISetToken _setToken, address _issuanceModule) external {
+    ComponentInvokeWrapData[] memory invokeUnwrapData = new ComponentInvokeWrapData[](_unwrapData.length);
+
+    for (uint256 i = 0; i < _unwrapData.length; ++i) {
+
+      invokeUnwrapData[i] = ComponentInvokeWrapData({
```
ckoopmann, line 430: Isn't this data the zero data for `ComponentInvokeWrapData` i.e. isn't this initialization unnecessary / wasting gas ?

ckoopmann (changes requested):
Using user provided calldata seems dangerous to me, especially when the calltarget is an erc20 token.
In general there seem to be a few places where we might need additional checks to make sure that attackers can't steal leftover transactions from previous issuance / redemption transactions.

Otherwise great work though 👍

contracts/exchangeIssuance/FlashMintWrapped.sol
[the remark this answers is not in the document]
ckoopmann: As mentioned in my review only verifying the callTarget and using user specified calldata seems dangerous to me. Especially if the user will be able to encode `approve` or `transfer` calls in this calldata.

contracts/exchangeIssuance/FlashMintWrapped.sol
ckoopmann: Some potential "attack vectors". (haven't really tested any of those or even thought them through to the end so feel free to challenge).

1. User issues 0 amount of set token and instead of the wrapping encodes an `approve` transaction. Afterwards he has the power to spend this contracts wrapped tokens.
2. When there is some underlying token left in the contract the user just wraps that and uses it to issue set token.

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,1087 @@
...
+      _swapToExact(
+        _inputToken, // input
+        IERC20(_swapData[i].underlyingERC20), // output
+        _swapData[i].buyUnderlyingAmount, // buy amount: must come from flash mint caller, we do not know the exchange rate wrapped <> unwrapped
+        _maxAmountInputToken, // maximum spend amount: _maxAmountInputToken as transferred by the flash mint caller
+        _swapData[i].dexData // dex path fees data etc.
+      );
+
+      uint256 underlyingBalance = IERC20(_swapData[i].underlyingERC20).balanceOf(address(this));
+      require(underlyingBalance >= _swapData[i].buyUnderlyingAmount, "FlashMint: UNDERBOUGHT_COMPONENT");
```
ckoopmann, line 813: This doesn't account for any underlying balance that might have been left over in the contract no ?

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,1087 @@
...
+      }
+
+      // transform wrapped token into unwrapped version (unless it's the same)
+      if (_swapData[i].underlyingERC20 != _redeemComponents[i]) {
+        // snapshot unwrapped balance before to compute the actual redeemed amount of underlying token (due to unknown exchange rate)
+        uint256 unwrappedBalanceBefore = IERC20(_swapData[i].underlyingERC20).balanceOf(
+          address(this)
+        );
+
+        _unwrapComponent(
```
ckoopmann, line 880: Wouldn't this allow the user to unwrap left over wrapped Token balance of the contract and extract the underlying tokens ?

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,1087 @@
...
+
+        // recompute actual redeemed amount to the underlyingERC20 token amount obtained through unwrapping
+        uint256 unwrappedBalanceAfter = IERC20(_swapData[i].underlyingERC20).balanceOf(
+          address(this)
+        );
+        redeemedAmount = unwrappedBalanceAfter.sub(unwrappedBalanceBefore);
+      }
+
+      // swap redeemed and unwrapped component to output token
+      uint256 boughtOutputTokenAmount = _swapFromExact(
```
ckoopmann, line 893: Also here I think we need a check to make sure the user doesn't sell more underlying tokens than he has previously obtained in unwrapping.

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,1087 @@
...
+        // no wrapping needed
+        continue;
+      }
+
+      // get the wrap adapter directly from the integration registry
+      IWrapV2Adapter wrapAdapter = IWrapV2Adapter(_getAndValidateAdapter(_wrapData[i].integrationName, _wrapModule));
+
+      // get wrap call info from adapter
+      (invokeWrapData[i].callTarget, invokeWrapData[i].value, invokeWrapData[i].callData) = wrapAdapter
+        .getWrapCallData(
```
ckoopmann, line 401: I think it'd be better to integrate this call into the issuance / redemption methods, since it would avoid many of the issues I mentioned elsewhere.

contracts/exchangeIssuance/FlashMintWrapped.sol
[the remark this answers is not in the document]
ckoopmann: The original implementation, calling the adapters directly in the issuance / redemption tx seems more robust and secure to me. Allowing the user to meddle with the calldata of chain opens various issues outlined below.

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: Setting the actual values in line 445 fails if the array element has not been initialized before, but it should work with just 
`invokeUnwrapData.length++`. Not sure if that saves gas, but it could. I can try that

Or did you have something else in mind?

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,1087 @@
...
+          _swapData[i].buyUnderlyingAmount,
+          _swapData[i].underlyingERC20,
+          _requiredComponents[i]
+        );
+      }
+
+      // ensure obtained component amount covers required component amount for issuance
+      // this is not already covered through _swapToExact because it does not take wrapping into consideration
+      uint256 componentBalanceAfter = IERC20(_requiredComponents[i]).balanceOf(address(this));
+      uint256 componentAmountObtained = componentBalanceAfter.sub(componentBalanceBefore);
```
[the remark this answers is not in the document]
bergben: the `swapToExact` should guarantee that exactly `buyUnderlyingAmount` is bought by the user. 
If the input amount is insufficient that call should fail AFAIK. 

In _wrapComponent, only the `buyUnderlyingAmount` is approved to the wrap call target, meaning no more than exactly what the user obtained can be spent for wrapping 

componentAmountObtained here makes sure that with this buyUnderlyingAmount -> wrap action, the user actually acquired enough component units to cover the issuance cost

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: Note that this underlyingBalance is only used here to fail early. 
But you are right, even for that it would have to take into account the previously present balance. 

I wasn't sure if it is worth the gas here though, especially when fixing this... It would just fail either at wrapping or at componentAmountObtained...

Would be very helpful to get your opinion on that

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: I had a discussion regarding this with Richard. I had it initially with direct calls to our wrap adapters as I think you have seen. 
callData is provided by the user, yes. But the callTarget is allowListed or a component of set tokens see line 980. AllowList is in control of the contract owner (IndexCoop), SetToken components can also only be added by IndexCoop.

contracts/exchangeIssuance/FlashMintWrapped.sol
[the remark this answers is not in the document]
bergben: @ckoopmann you do have a point. 

I think most of the things now discussed in this PR come down to how important it is to protect the potential left over dust in this contract... There shouldn't be any other funds in this contract? 

Seems like if we want to fight to keep this dust in this contract we need additional checks and we probably end up with the same gas cost as using the wrap adapters directly 

If we don't add those checks we put security at risk because of potential unknowns

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: Good points @ckoopmann, really appreciate your thinking about potential attack vectors here. 
Let's discuss in the other comment regarding if it's better to go back to wrap adapters...

contracts/exchangeIssuance/FlashMintWrapped.sol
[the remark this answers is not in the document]
bergben: @FlattestWhite what's your opinion on this after @ckoopmann inputs?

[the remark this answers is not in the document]
bergben: Resolving because discussed in other places

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: Implemented as discussed

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: I am resolving this conversation, the decision regarding BIMv2 is not to be made in this PR. 
I renamed debtIssuanceModule to issuanceModule. BIMv2 would be compatible with IDebtIssuanceModule so all options are open

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: I removed this in the newest version, I don't think it adds much value as is, it would require another snapshot which I think is not worth it since it will fail in the next step anyway

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: resolved with direct use of wrap adapters because we use the actual redeemedAmount as input for unwrapping amount

bergben:
@FlattestWhite  @ckoopmann FYI
This are the results for transaction fee comparison (in Gwei on localhost, not accurate but should give an idea):

 #issueExactSetFromETH: paying with ETH
1097542 Direct wrap adapters
1069513 Wrap call data

 #issueExactSetFromERC20: paying with USDC
1121616 Direct wrap adapters
1093689 Wrap call data

 #issueExactSetFromERC20: paying with DAI
1130365 Direct wrap adapters
1102438 Wrap call data

 #redeemExactSetForETH: receiving ETH
1016465 Direct wrap adapters
  982453 Wrap call data

 #redeemExactSetForERC20: receiving USDC
996310 Direct wrap adapters
962363 Wrap call data

 #redeemExactSetForERC20: receiving DAI
1055592 Direct wrap adapters
1021645 Wrap call data

ckoopmann (approved):
LGTM! Also thanks for the detailed gas analysis 👍

test/integration/ethereum/flashMintWrappedIntegration.spec.ts
```diff
@@ -0,0 +1,695 @@
...
+                      : await outputToken.balanceOf(owner.address);
+
+                  let outputObtained = outputBalanceAfter.sub(outputBalanceBefore);
+
+                  if (tokenName == "ETH") {
+                    const transactionFee = await getTxFee(tx);
+                    outputObtained = outputObtained.add(transactionFee);
+                  }
+
+                  expect(outputObtained.gte(minAmounOutput)).to.be.true;
```
FlattestWhite, line 653:
```suggestion
                  expect(outputObtained.gte(minAmoungOutput)).to.be.true;
```

FlattestWhite (approved):
had one question but other than that lgtm

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,992 @@
...
-    using PreciseUnitMath for uint256;
-    using SafeERC20 for IERC20;
-    using SafeERC20 for ISetToken;
-
-    struct WrappedCalldata {
-        address target;
-        uint256 quantity;
-        bytes callData;
-    }
+contract FlashMintWrapped is ReentrancyGuard, Withdrawable {
```
FlattestWhite, line 55: `Withdrawable` seems to be imported from `external/aavev2/utils` which doesn't make sense as an import for this contract. I think we can either move that out to a shared utils or just use what @ckoopmann did in ExchangeIssuanceZeroEx 
```
    /**
     * Withdraw slippage to selected address
     *
     * @param _tokens    Addresses of tokens to withdraw, specifiy ETH_ADDRESS to withdraw ETH
     * @param _to        Address to send the tokens to
     */
    function withdrawTokens(IERC20[] calldata _tokens, address payable _to) external onlyOwner payable {
        for(uint256 i = 0; i < _tokens.length; i++) {
            if(address(_tokens[i]) == ETH_ADDRESS){
                _to.sendValue(address(this).balance);
            }
            else{
                _tokens[i].safeTransfer(_to, _tokens[i].balanceOf(address(this)));
            }
        }
    }
```

FlattestWhite (approved):

contracts/exchangeIssuance/DEXAdapter.sol
```diff
@@ -682,7 +682,10 @@ library DEXAdapter {
         returns (uint256)
     {
         _safeApprove(IERC20(_path[0]), address(_router), _amountIn);
-        return _router.swapExactTokensForTokens(_amountIn, _minAmountOut, _path, address(this), block.timestamp)[1];
+        uint256[] memory result = _router.swapExactTokensForTokens(_amountIn, _minAmountOut, _path, address(this), block.timestamp);
+        // result = uint[] memory	The input token amount and all subsequent output token amounts.
+        // we are usually only interested in the actual amount of the output token (so result element at the last place)
+        return result[result.length-1];
```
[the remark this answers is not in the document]
bergben: you mean the contracts that use / include this library in their deployment? 

Potentially, I asked in the discord channel. 
It is only a potential problem in `ExchangeIssuanceLeveraged` and `FlashMintLeveragedForCompound`.
I thought @ckoopmann or @ElmoArmy might be better suited to establish any potential risks there...

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,992 @@
...
+    IWrapV2Adapter _wrapAdapter = IWrapV2Adapter(_getAndValidateAdapter(_unwrapData.integrationName));
+
+    // 2. get unwrap call info from adapter
+    (address _wrapCallTarget, uint256 _wrapCallValue, bytes memory _wrapCallData) = _wrapAdapter
+      .getUnwrapCallData(
+        _underlyingToken,
+        _wrappedToken,
+        _unwrapAmount,
+        address(this),
+        _unwrapData.wrapData
```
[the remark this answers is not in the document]
bergben: hmm I was focused on Aave and Compound it seems where it is not needed
Also for Yearn it wouldn't be needed. Those are the protocols we had wrap adapters for in the past according to the repo. 
Do you think it makes sense to add it to really ensure generic usability of this contract? 
TBH I don't know a protocol that would require this...

contracts/exchangeIssuance/DEXAdapter.sol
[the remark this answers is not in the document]
FlattestWhite: Yeah I was referring to the contracts that use this library - maybe it's fine and add a note to this contract that this library was updated on this date with this change for this reason.

contracts/exchangeIssuance/FlashMintWrapped.sol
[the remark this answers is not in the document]
FlattestWhite: Nah it's all good, if it's not needed. I was just questioning since on the other side, we run the approvals

contracts/exchangeIssuance/FlashMintWrapped.sol
bergben: done

test/integration/ethereum/flashMintWrappedIntegration.spec.ts
bergben: fixed

utils/deploys/deployExtensions.ts
```diff
@@ -42,6 +42,8 @@ import { GIMExtension__factory } from "../../typechain/factories/GIMExtension__f
 import { GovernanceExtension__factory } from "../../typechain/factories/GovernanceExtension__factory";
 import { StreamingFeeSplitExtension__factory } from "../../typechain/factories/StreamingFeeSplitExtension__factory";
 import { WrapExtension__factory } from "../../typechain/factories/WrapExtension__factory";
+import { FlashMintWrapped__factory } from "@typechain/factories/FlashMintWrapped__factory";
```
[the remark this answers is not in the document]
bergben: fixed

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,992 @@
...
+   * @param _issuanceModule            IDebtIssuanceModule used to issue and redeem tokens
+   * @param _wrapModule                WrapModuleV2 used to obtain a valid wrap adapter
+   */
+  constructor(
+    DEXAdapter.Addresses memory _dexAddresses,
+    IController _setController,
+    IDebtIssuanceModule _issuanceModule,
+    address _wrapModule
+  ) public {
+    setController = _setController;
```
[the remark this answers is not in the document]
bergben: fixed

contracts/exchangeIssuance/FlashMintWrapped.sol
```diff
@@ -17,220 +17,992 @@
...
-import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
-import { SafeMath } from "@openzeppelin/contracts/math/SafeMath.sol";
-import { ReentrancyGuard } from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
-
-import { IController } from "../interfaces/IController.sol";
-import { IDebtIssuanceModule } from "../interfaces/IDebtIssuanceModule.sol";
-import { ISetToken } from "../interfaces/ISetToken.sol";
-import { IWETH } from "../interfaces/IWETH.sol";
-import { PreciseUnitMath } from "../lib/PreciseUnitMath.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
```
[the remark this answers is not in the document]
bergben: adjusted configuration

contracts/exchangeIssuance/DEXAdapter.sol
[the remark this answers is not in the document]
bergben: added a note

bergben:
@FlattestWhite fixed everything you pointed out and added test cases for 
- Set token with only unwrapped components
- Set token with mixed components (cDAI, USDC, cUSDT)

PR #120, opened 2022-11-14
StakeWise Reinvestment Extension

FlattestWhite:
index is currently building the wsETH2 token, a SetToken that wraps the StakeWise sETH2 token and automatically reinvests accured rETH2 back into sETH2. To do this, we're running a keeper that periodically calls this extension contract which uses the AirdropModule to first absorb the rETH2 into the Set before trading it into sETH2 in one function call.


external/abi/set/TradeModule.json
```diff
@@ -0,0 +1,10 @@
+{
```
ckoopmann, line 1: I assume that this is our own deployment of `TradeModule.json`. In that case maybe we should put it in `external/abi/index-protocol` or something like that to keep abis of the original and fork deployments separate.

test/integration/ethereum/stakeWiseReinvestmentExtensionIntegration.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+      it("should set the correct executionSettings", async () => {
+        expect(await extension.settings()).to.deep.eq([
+          "UniswapV3ExchangeAdapterV2",
+          "0x20bc832ca081b91433ff6c17f85701b6e92486c50001f4fe2e637202056d30016725477c5da089ab0a043a01",
+        ]);
+      });
+    });
+
+    describe("#initialize", () => {
+      async function subject() {
```
ckoopmann, line 144: For completeness I'd add some tests that assert that unauthorized callers cannot call protected methods.

contracts/adapters/StakeWiseReinvestmentExtension.sol
```diff
@@ -0,0 +1,137 @@
...
+    /* ========== Structs ================= */
+
+    struct ExecutionSettings {
+        string exchangeName;
+        bytes exchangeCallData;
+    }
+
+    /* ========== State Variables ========= */
+
+    ISetToken public immutable setToken;                // The set token
```
ckoopmann, line 57: I don't think these comments add much additional information. I suggest to either extend them or remove them.

contracts/adapters/StakeWiseReinvestmentExtension.sol
```diff
@@ -0,0 +1,137 @@
...
+    using SafeCast for int256;
+
+    /* ============ Constants ============= */
+
+    address public constant S_ETH2 = 0xFe2e637202056d30016725477c5da089Ab0A043A;
+    address public constant R_ETH2 = 0x20BC832ca081b91433ff6c17f85701B6e92486c5;
+
+    /* ========== Structs ================= */
+
+    struct ExecutionSettings {
```
ckoopmann, line 50: Maybe add short comment explaining this struct and each field ror maximum readability.

contracts/adapters/StakeWiseReinvestmentExtension.sol
```diff
@@ -0,0 +1,137 @@
...
+    IAirdropModule public immutable airdropModule;      // The airdrop module
+    ITradeModule public immutable tradeModule;          // The trade module
+    ExecutionSettings public settings;                  // The execution settings
+
+    /* ============  Constructor ============ */ 
+    /**
+     * Sets state variables
+     * 
+     * @param _manager Manager contract
+     * @param _airdropModule Airdrop module contract
```
ckoopmann, line 67: These parameter explanations don't add much information. I suggest extending at least some of them. For example for the airdrop module a short explanation why we need it here / what it is used for would be nice.

ckoopmann (commented):
LGTM! Just some minor comments mostly regarding documentation / code comments.

Btw: If you happen to know a way to combine the coverage from our integration tests into the overall report that'd be amazing. Otherwise our reported coverage will keep going down when we do most of our testing in integration tests.

ckoopmann (approved):
LGTM! Great work.

PR #122, opened 2022-11-29
FIXED rebalance extension

ckoopmann:
Extension / Adapter to rebalance FIXED products and maintain target allocation among fCash maturities.


utils/constants.ts
```diff
@@ -18,6 +18,7 @@ export const MAX_INT_256 = "0x7fffffffffffffffffffffffffffffffffffffffffffffffff
 export const MIN_INT_256 = "-0x8000000000000000000000000000000000000000000000000000000000000000";
 export const ONE_DAY_IN_SECONDS = BigNumber.from(60 * 60 * 24);
 export const ONE_HOUR_IN_SECONDS = BigNumber.from(60 * 60);
+export const ONE_MONTH_IN_SECONDS = ONE_DAY_IN_SECONDS.mul(30);
 export const ONE_YEAR_IN_SECONDS = BigNumber.from(31557600);
```
jeffywu, line 22: FYI, Notional uses 360 days not 365. I'm not sure where this constant is used but wanted to call that out.

ckoopmann: I'm not using this constant in any of my notional related test, but great point. Thanks for pointing this out.🙏

FlattestWhite (commented):
Before making a detailed review of this PR, I've got a few questions to answer:
1. How do you envision this contract to fit into our existing keeper system. Currently, I've got this lambda that continuously calls the extension contract method shouldRebalance() and depending on the response either calls rebalance(), iterateRebalance() or ripcord(). The difference with them is that the caller only needs to specify the ExchangeAdapter that they need to be called with. How are _share and _minPositions calculated by the caller?
2. How often should rebalance be called? Is a cooldown period?
3. How does a keeper know not to call rebalance?

ckoopmann:
> > Before making a detailed review of this PR, I've got a few questions to answer:
> > 
> > 1. How do you envision this contract to fit into our existing keeper system. Currently, I've got this lambda that continuously calls the extension contract method shouldRebalance() and depending on the response either calls rebalance(), iterateRebalance() or ripcord(). The difference with them is that the caller only needs to specify the ExchangeAdapter that they need to be called with. How are _share and _minPositions calculated by the caller?
> > 2. How often should rebalance be called? Is a cooldown period?
> > 3. How does a keeper know not to call rebalance?
> 
> From my understanding we will only call these rebalance methods when one of the underlying fcash tokens matures. (so every three months). So I don't think we need to necessarily integrate this into the keeper system at all but can call it manually once every three months. (the name "RebalanceExtension" might be a bit misleading here since we are not planning to do the kind of continous rebalancing that we do for some of our other products - although the contract would support rebalancing at any point)
> 
> `_minPositions` can be calculated by simulating the rebalance call checking the positions afterwards and then subtracting a few percentage points. This is basically just for frontrunning protection.
> 
> `_share` allows us to split the rebalance trade over multiple transactions to limit price impact. I'm not sure if this is an issue at all and either way wont be one until we reach significant AUM. So for the beginning we can probably just set it to "1.0".

Note that if we at some point did decide to do "continous rebalancing" (which we currently don't plan to do). Then the keeper could determine wether rebalancing is needed by calling `getUnderweightPositions`, and using a certain minimum threshold. (since there will always be some minor deviation from the target allocations due to fluctuations in interest rates).

test/integration/ethereum/fixedRebalanceExtension.spec.ts
```diff
@@ -0,0 +1,491 @@
...
+          );
+          await baseManagerV2.connect(operator).addExtension(rebalanceExtension.address);
+          currencyId = await rebalanceExtension.currencyId();
+        });
+
+        describe("#getAbsoluteMaturities", () => {
+          function subject() {
+            return rebalanceExtension.getAbsoluteMaturities();
+          }
+          it("should work", async () => {
```
FlattestWhite, line 157:
```suggestion
          it("should have the same members as componentMaturities", async () => {
```

test/integration/ethereum/fixedRebalanceExtension.spec.ts
```diff
@@ -0,0 +1,491 @@
...
+          let caller: Signer;
+          beforeEach(() => {
+            subjectMaturities = [ONE_MONTH_IN_SECONDS.mul(3)];
+          });
+
+          function subject() {
+            return rebalanceExtension.connect(caller).setValidMaturities(subjectMaturities);
+          }
+
+          describe("when the caller is not the owner", () => {
```
FlattestWhite, line 174:
```suggestion
          describe("when the caller is not the operator", () => {
```

FlattestWhite (approved):
just to confirm - it looks like we'd need to deploy an extension contract per FIXED product that i.e. FIXED-DAI and FIXED-USDC will each need their own extension contract

ckoopmann:
> just to confirm - it looks like we'd need to deploy an extension contract per FIXED product that i.e. FIXED-DAI and FIXED-USDC will each need their own extension contract

Yes, we would deploy one instance for each FIXED product. I assumed this was standard for our extensions. (Please correct me if i'm wrong). 

Note that the smart contract code should work for any FIXED product. I will extend the tests to run for all of our fixed products to make sure that is actually the case.

contracts/adapters/FixedRebalanceExtension.sol
```diff
@@ -0,0 +1,483 @@
...
+    ISetToken setToken;                                      // Instance of leverage token
+    INotionalTradeModule notionalTradeModule;                // Instance of leverage token
+    INotionalProxy notionalProxy;                            
+    IWrappedfCashFactory wrappedfCashFactory;                // Instance of leverage token
+    address underlyingToken;
+    address assetToken;
+    uint256[] maturities;                                    // Array of relative maturities in seconds (i.e. 3 months / 6 months)
+    uint256[] allocations;                                   // Relative allocations 
+    uint16 public currencyId;
+    bool tradeViaUnderlying;
```
ElmoArmy, line 77:
```suggestion
    // /* ============ State Variables ============ */

    mapping(uint256 => bool) internal isValidMaturity;           // Mapping of valid maturities
    uint256[] internal validMaturities;                          // Array of valid maturities
    uint256[] internal maturities;                               // Array of relative maturities in seconds (i.e. 3 months / 6 months)
    uint256[] internal allocations;                              // Relative allocations 

    ISetToken internal immutable setToken;                       // Instance of leverage token
    INotionalTradeModule internal immutable notionalTradeModule; // Instance of leverage token
    INotionalProxy internal immutable notionalProxy;                            
    IWrappedfCashFactory internal immutable wrappedfCashFactory; // Instance of leverage token
    address internal immutable underlyingToken;
    address internal immutable assetToken;
    uint16 public immutable currencyId;
    
    bool internal tradeViaUnderlying;
```

I think some of these could be better to have with public visibility, but this is the less controversial change.

ElmoArmy (changes requested):
### Testing
- Overall consistent and I like what @FlattestWhite has suggested. When we read the output from a test it should make sense without **much** context. 
- Further more on naming, I understand that you created the context of some tests by the name of the function and the test names as "should work". Being more descriptive on these names would help others understand the context in which a user would care to call the function or how it is crucial to user requirement. I recognize this as a nit for development but these tests provide such awesome documentation to power users/devs and should strive for as good as possible. 

### Coverage
I think 100% test coverage is a bad requirement but we should make an effort to understand the coverage our tests provide and have it guide discussions.

``# I utilized this command to get the following results. All tests passed.``
``
COVERAGE=true INTEGRATIONTEST=true FORK=true VERBOSE=true node --max-old-space-size=4096 ./node_modules/.bin/hardhat coverage --testfiles test/integration/ethereum/fixedRebalanceExtension.spec.ts``


File                                     |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines 
-----------------------------------------|----------|----------|----------|----------|----------------|
 FixedRebalanceExtension.sol   |    93.02 |     62.5 |    92.59 |    93.02 | 416,417,420 
 
-  `_getFCashComponentsAndMaturities` and `_absoluteToRelativeMaturity` are internal functions of the extension that are not covered by these tests. Removing them, increased all but branch coverage to 100%. The compiler should remove these functions from the deployed bytecode if they are never referenced. I suggest being more explicit and removing the unused code.

- Testing every branch (especially in internal functions) is not super important to me, but understanding the gap is important. As a general rule, the `else case` of a require statement is already well tested (the call reverts). Branches caused by `if` statements are harder to assess for criticality of appropriate coverage. 
  - This Extension has two `if` branches not covered by existing tests, Lines 219 and 432. Make sure that the else paths are reachable and if reached, the expected behavior would still take place.
- "access control" require statements and modifiers are worth having coverage for integration tests but more focused on the state changes that might affect access. For example, we might want to test us changing the operator and ensuring that the previous operator could no longer take privileged actions. Looks like this contract makes use of the onlyOperator modifier that presumably has been tested elsewhere, so we can assume it works but I still see it as such a critical path that we should make certain that this path works as expected. 

- "Invalid state" require statements, I think merit integration testing if it is reasonable that such branch would be taken or if removing the condition would be catastrophic for assumptions further in the code.  
  - Take for example:
`function rebalance(uint256 _share, uint256[] memory _minPositions)`
and the require statements
`require(_share > 0, "Share must be greater than 0");`
`require(_share <= 1 ether, "Share cannot exceed 100%");`
   In my testing removing these statements did not make any test fail and these calls are restricted to a privileged account. I therefore wouldn't worry about coverage for these cases.
 
 - All branches that are not covered seem to be internal and none of them are exposed through external functions accessible by unprivileged account. So no need to block on branch coverage. 

### Security
- Seems like the only security assumptions that exist are that of a malicious operator due to all state changing functions being restricted. Unless there is a requirement to be less restrictive on who can modify the extension's state, I think this is appropriately secure enough. Further locking down the operator to reasonable checks would be great though. Is there not a way to get the valid maturities/ allocations from the notional protocol given the setToken passed in?

- We might want to consider reentrancy in the external functions. At a glance though and with my understanding of the Set Protocol, we shouldn't need to worry about that. 

- We should be explicit about visibility for state variables. Most seem to be internal which is a good default, but I think some of these would make sense to be public and have the default getters exposed. Furthermore, it seems like almost all the values passed in to the constructor are immutably set to storage and if declared as immutable could be more efficient.

contracts/adapters/FixedRebalanceExtension.sol
```diff
@@ -0,0 +1,483 @@
...
+ *
+ * Smart contract that enables rebalancing of FIXED products
+ * Will sell redeem fCash positions and mint underweight fCash positions via NotionalTradeModule
+ */
+contract FixedRebalanceExtension is BaseExtension {
+    using PreciseUnitMath for uint256;
+    using SafeMath for uint256;
+
+
+    struct CallArguments {
```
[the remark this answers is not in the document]
ckoopmann: Well spotted. Removed it

contracts/adapters/FixedRebalanceExtension.sol
```diff
@@ -0,0 +1,483 @@
...
+        // TODO: Verify that this is not an issue
+        uint256 remainingMaturity = _absoluteMaturity.sub(block.timestamp);
+        // Valid maturities is sorted in ascending order
+        for(uint256 i = 0; i < validMaturities.length; i++) {
+            if(validMaturities[i] >= remainingMaturity) {
+                return validMaturities[i];
+            }
+        }
+        revert("Remaining maturity is larger than any valid maturity");
+    }
```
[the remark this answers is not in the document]
ckoopmann: Well spotted. Removed it

contracts/adapters/FixedRebalanceExtension.sol
```diff
@@ -0,0 +1,483 @@
...
+    returns (address[] memory fCashComponents, uint256[] memory fCashMaturities)
+    {
+        fCashComponents = notionalTradeModule.getFCashComponents(setToken);
+        fCashMaturities = new uint256[](fCashComponents.length);
+        for(uint256 i = 0; i < fCashComponents.length; i++) {
+            fCashMaturities[i] = _absoluteToRelativeMaturity(
+                IWrappedfCashComplete(fCashComponents[i]).getMaturity()
+            );
+        }
+    }
```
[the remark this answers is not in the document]
ckoopmann: Well spotted. Removed it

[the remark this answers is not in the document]
ckoopmann: Well spotted. Removed it

ckoopmann:
> ### Testing
> * Overall consistent and I like what @FlattestWhite has suggested. When we read the output from a test it should make sense without **much** context.
> * Further more on naming, I understand that you created the context of some tests by the name of the function and the test names as "should work". Being more descriptive on these names would help others understand the context in which a user would care to call the function or how it is crucial to user requirement. I recognize this as a nit for development but these tests provide such awesome documentation to power users/devs and should strive for as good as possible.

I agree "should work" is a terrible test name. I replaced all of its occurences with more informative names.

> 
> ### Coverage
> I think 100% test coverage is a bad requirement but we should make an effort to understand the coverage our tests provide and have it guide discussions.
> 
> `# I utilized this command to get the following results. All tests passed.` ` COVERAGE=true INTEGRATIONTEST=true FORK=true VERBOSE=true node --max-old-space-size=4096 ./node_modules/.bin/hardhat coverage --testfiles test/integration/ethereum/fixedRebalanceExtension.spec.ts`
> 
> File	% Stmts	% Branch	% Funcs	% Lines	Uncovered Lines
> FixedRebalanceExtension.sol	93.02	62.5	92.59	93.02	416,417,420
> * `_getFCashComponentsAndMaturities` and `_absoluteToRelativeMaturity` are internal functions of the extension that are not covered by these tests. Removing them, increased all but branch coverage to 100%. The compiler should remove these functions from the deployed bytecode if they are never referenced. I suggest being more explicit and removing the unused code.
> * Testing every branch (especially in internal functions) is not super important to me, but understanding the gap is important. As a general rule, the `else case` of a require statement is already well tested (the call reverts). Branches caused by `if` statements are harder to assess for criticality of appropriate coverage.
>   

I removed the unused methods and reached 100% line coverage. I looked at the missed branches added some extra coverage where it was worth the effort, but I agree we don't necessarily need 100% branch coverage. 

Also I tried to add a CI step that reports the coverage from the integration test but it ended up overwriting the coverage data from the unit tests so I removed it again (see: https://github.com/IndexCoop/index-coop-smart-contracts/pull/122/commits/a7a8c0cba1d8a9da32b070db9488fb4f496cf261) . Would be great if we could find a way to aggregate those results somehow. 

>   * This Extension has two `if` branches not covered by existing tests, Lines 219 and 432. Make sure that the else paths are reachable and if reached, the expected behavior would still take place.
> * "access control" require statements and modifiers are worth having coverage for integration tests but more focused on the state changes that might affect access. For example, we might want to test us changing the operator and ensuring that the previous operator could no longer take privileged actions. Looks like this contract makes use of the onlyOperator modifier that presumably has been tested elsewhere, so we can assume it works but I still see it as such a critical path that we should make certain that this path works as expected.


> * "Invalid state" require statements, I think merit integration testing if it is reasonable that such branch would be taken or if removing the condition would be catastrophic for assumptions further in the code.
>   
>   * Take for example:
>     `function rebalance(uint256 _share, uint256[] memory _minPositions)`
>     and the require statements
>     `require(_share > 0, "Share must be greater than 0");`
>     `require(_share <= 1 ether, "Share cannot exceed 100%");`
>     In my testing removing these statements did not make any test fail and these calls are restricted to a privileged account. I therefore wouldn't worry about coverage for these cases.
> * All branches that are not covered seem to be internal and none of them are exposed through external functions accessible by unprivileged account. So no need to block on branch coverage.
> 
> ### Security
> * Seems like the only security assumptions that exist are that of a malicious operator due to all state changing functions being restricted. Unless there is a requirement to be less restrictive on who can modify the extension's state, I think this is appropriately secure enough. Further locking down the operator to reasonable checks would be great though. Is there not a way to get the valid maturities/ allocations from the notional protocol given the setToken passed in?
> * We might want to consider reentrancy in the external functions. At a glance though and with my understanding of the Set Protocol, we shouldn't need to worry about that.
> * We should be explicit about visibility for state variables. Most seem to be internal which is a good default, but I think some of these would make sense to be public and have the default getters exposed. Furthermore, it seems like almost all the values passed in to the constructor are immutably set to storage and if declared as immutable could be more efficient.

Agree with all of your points and just implemented most of your suggestions.
Regarding the security issues we might want to review some of them (such as reentrancy) if we decide to open up some of those methods to be called by anyone.

contracts/adapters/FixedRebalanceExtension.sol
```diff
@@ -47,34 +47,25 @@ contract FixedRebalanceExtension is BaseExtension {
...
-    address underlyingToken;
-    address assetToken;
-    uint256[] maturities;                                    // Array of relative maturities in seconds (i.e. 3 months / 6 months)
-    uint256[] allocations;                                   // Relative allocations 
-    uint16 public currencyId;
-    bool tradeViaUnderlying;
+    uint256[] internal maturities;                               // Array of relative maturities in seconds (i.e. 3 months / 6 months)
+    uint256[] internal allocations;                              // Relative allocations 
+
+    ISetToken internal immutable setToken;                       // Instance of leverage token
```
ElmoArmy, line 60: I'd consider making this public so we don't have to look at constructor args to know what Product this is for.

contracts/adapters/FixedRebalanceExtension.sol
```diff
@@ -47,34 +47,25 @@ contract FixedRebalanceExtension is BaseExtension {
...
-        uint40 maturity;
-        uint256 minMintAmount;
-        address sendToken;
-        uint256 sendAmount;
-    }
-
     // /* ============ Events ============ */
 
     event AllocationsUpdated(uint256[] _maturities, uint256[] _allocations);
     event MaturityStatusSet(uint256 _maturity, bool _newStatus);
```
ElmoArmy, line 53: This event is no longer referenced with your refactor.
```suggestion
```

ElmoArmy (changes requested):
Left over event (MaturityStatusSet) should probably be removed after your refactor. Another poke about setting some of the state variables as public, but I don't feel strongly about it. Great refactor.

contracts/adapters/FixedRebalanceExtension.sol
ckoopmann: Good point. Done

ElmoArmy (approved):

PR #124, opened 2023-01-02
(feat): Add SetTokenRateViewer contract

FlattestWhite:


contracts/viewers/SetTokenRateViewer.sol
```diff
@@ -0,0 +1,68 @@
...
+        component = _component;
+    }
+
+    /* =========== External Functions ============ */
+
+    /**
+     * @return the amount of component per set token.
+     */
+    function getRate() external view override returns (uint256) {
+        return abs(setToken.getTotalComponentRealUnits(address(component)));
```
ckoopmann, line 57: Do we actually want to support negative component units ? (by just taking the absolute value)
It seems a bit strange to me.
If not I suggest throwing an exception instead of just flipping the sign.

package.json
```diff
@@ -42,9 +42,9 @@
     "test": "npx hardhat test --network localhost",
     "test:fast": "TS_NODE_TRANSPILE_ONLY=1 npx hardhat test --network localhost --no-compile",
     "test:fast:compile": "TS_NODE_TRANSPILE_ONLY=1 npx hardhat test --network localhost",
-    "test:integration:polygon": "INTEGRATIONTEST=true VERBOSE=true yarn run test test/integration/polygon/**",
-    "test:integration:optimism": "INTEGRATIONTEST=true VERBOSE=true yarn run test test/integration/optimism/**",
-    "test:integration:ethereum": "INTEGRATIONTEST=true VERBOSE=true yarn run test test/integration/ethereum/**",
+    "test:integration:polygon": "ls test/integration/polygon; INTEGRATIONTEST=true VERBOSE=true yarn run test test/integration/polygon/**/*.spec.ts",
```
ckoopmann: Do we want to keep the ls commands here ? if so I suggest adding it to the ethereum command as well for consistency.

test/integration/ethereum/viewer/setTokenRateViewerIntegration.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+import { Account } from "@utils/types";
+import DeployHelper from "@utils/deploys";
+import { getAccounts, getWaffleExpect, ether } from "@utils/index";
+
+import { IERC20, ISetToken, SetTokenRateViewer } from "../../../../typechain";
+import { PRODUCTION_ADDRESSES, STAGING_ADDRESSES } from "../addresses";
+
+const expect = getWaffleExpect();
+
+if (process.env.INTEGRATIONTEST) {
```
ckoopmann, line 12: Is there a case here where we test the logic when the position is negative ? if not I suggest adding it.

ckoopmann (commented):
Mostly LGTM, just one or two minor questions. Don't forget to get the CI green before merging as well. (Seems like the file pattern for the integration tests does not work).

package.json
ckoopmann: Seems like this was removed while I was reviewing.

contracts/viewers/SetTokenRateViewer.sol
FlattestWhite: Yeah I wasn't sure what to do about negative component units. Originally I had it as just throwing, so just reverted to that

test/integration/ethereum/viewer/setTokenRateViewerIntegration.spec.ts
FlattestWhite: Yep there's a case for negative positions

FlattestWhite:
> Mostly LGTM, just one or two minor questions. Don't forget to get the CI green before merging as well. (Seems like the file pattern for the integration tests does not work).

yeah seems like CI doesn't like the globbing pattern

ckoopmann (approved):
Great work!

test/integration/ethereum/viewer/setTokenRateViewerIntegration.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    context("wsEth2", async () => {
+      it("should get the amount of sETH2 per wsETH2 token", async () => {
+        setToken = (await ethers.getContractAt("ISetToken", addresses.tokens.wsETH2)) as ISetToken;
+        component = (await ethers.getContractAt("IERC20", addresses.tokens.sETH2)) as IERC20;
+
+        await subject();
+
+        const rate = await viewer.getRate();
+        expect(rate).to.eq(ether(1));
```
edkim, line 48: @FlattestWhite do we also want to add an integration test for `getRate() > 1` _after_ performing a reinvestment of rETH2 to sETH2?

PR #125, opened 2023-01-02
Extended `minPositions` configuration to potentially allow opening up access to `rebalance` method

ckoopmann:
This PR splits up the configuration of the minimum positions to reach after the rebalance call into two parts:

`_rebalanceMinPositions`: Values passed by the caller to `rebalance` to provide slippage protection.
`minPositions`: State variable controlled by the "operator" used to verify lower bound of `_rebalanceMinPositions`. Each element in `_rebalanceMinPositions` has to be greater or equal to the weighted (by `_share`) average between the operator specified minPosition and the position prior to the rebalance call.

Note that when `minPositions` is specified to all zeroes the functionality should remain largely unchanged and the caller can specify arbitrary `_rebalanceMinPositions`.

Sensibly configuring  `minPositions` should provide an additional layer of security against malicious calls of `rebalance` (for example if private key of another 'allowedCaller' is compromised).
It might even enable making the `rebalance` method publicly callable, although this should be decided on with extreme caution.

### DIFF
--- a/contracts/adapters/FixedRebalanceExtension.sol
+++ b/contracts/adapters/FixedRebalanceExtension.sol
@@ -49,12 +49,13 @@ contract FixedRebalanceExtension is BaseExtension {
 
     // /* ============ Events ============ */
 
-    event AllocationsUpdated(uint256[] _maturities, uint256[] _allocations);
+    event AllocationsUpdated(uint256[] _maturities, uint256[] _allocations, uint256[] _minPositions);
 
     // /* ============ State Variables ============ */
 
     uint256[] internal maturities;                               // Array of relative maturities in seconds (i.e. 3 months / 6 months)
     uint256[] internal allocations;                              // Relative allocations 
+    uint256[] internal minPositions;                             // Minimum positions to achieve after rebalancing
 
     ISetToken public immutable setToken;                       
     INotionalTradeModule internal immutable notionalTradeModule; 
@@ -77,7 +78,8 @@ contract FixedRebalanceExtension is BaseExtension {
         address _underlyingToken,
         address _assetToken,
         uint256[] memory _maturities,
-        uint256[] memory _allocations
+        uint256[] memory _allocations,
+        uint256[] memory _minPositions
     )
         public
         BaseExtension(_manager)
@@ -90,7 +92,7 @@ contract FixedRebalanceExtension is BaseExtension {
         assetToken = _assetToken;
         currencyId = _notionalProxy.getCurrencyId(_assetToken);
 
-        _setAllocations(_maturities, _allocations);
+        _setAllocations(_maturities, _allocations, _minPositions);
     }
 
     // /* ============ External Functions ============ */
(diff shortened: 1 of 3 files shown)


contracts/adapters/FixedRebalanceExtension.sol
```diff
@@ -110,23 +112,24 @@ contract FixedRebalanceExtension is BaseExtension {
...
      * ONLY ALLOWED CALLER: Rebalances the positions towards the configured allocation percentages.
      *
-     * @param _share                Relative share of the necessary trade volume to execute (allows for splitting the rebalance over multiple transactions
-     * @param _minPositions         Minimum positions (in set token units) for each maturity after rebalance. (slippage protection)
+     * @param _share                 Relative share of the necessary trade volume to execute (allows for splitting the rebalance over multiple transactions
+     * @param _rebalanceMinPositions Minimum positions (in set token units) for each maturity after this rebalance operation. 
+     * @dev Will revert if _rebalanceMinPositions is lower than the minPositions configured by the operator (weighted by share)
      */
-    function rebalance(uint256 _share, uint256[] memory _minPositions) external onlyAllowedCaller(msg.sender) {
+    function rebalance(uint256 _share, uint256[] memory _rebalanceMinPositions) external onlyAllowedCaller(msg.sender) {
```
FlattestWhite, line 126: For _rebalanceMinPositions, are we able to derive that from existing storage so that we don't need the caller to pass it in? Ideally, if we were to open rebalance up to being called by anyone, then it shouldn't have any parameters. I guess if we could set _share and _rebalanceMinPositions, as storage variables, which can only be set by operator, then we'd can open rebalance up to being called by anyone? Such as chainlink keepers?

ckoopmann: Yes we can derive the minimum from a storage variable and we already do that. (see here:
https://github.com/IndexCoop/index-coop-smart-contracts/pull/125/files#diff-4e84502ce034d717a90159b1ab768a5f6eeafcb03e351853b5b1064675789211R239)

The idea here is that we use that storage variable as protection against malicious `rebalance` calls but then let the caller "overwrite" these values with higher ones to get the optimal slippage protection based on market conditions. 

Also in this current way we are quite flexible, for example if we set the `minPositions` to all zeroes then we basically keep the existing logic of letting the caller set arbitrary minPositions. (of course in that case we can't open it up to public calls).


I'm not sure if this might be a bit of over-engineering. But I thought if we just let the operator set these values then we have the tradeoff between either setting these values lower and paying a lot in slippage / sandwhich mev or setting them quite high but then having to continously update them to account for changing market conditions.


Does that make sense ? What do you think ?  @FlattestWhite

ckoopmann: I also tried to summarize this idea with different wording in the description / summary comment of the PR.

FlattestWhite (approved):

PR #128, opened 2023-01-19
Update readme and remove optimism from ci and configs

edkim:
This PR combines updated README with a clean up of Optimism references since CI was failing on OP integration tests. I figured since we don't support Optimism anymore, it's best to remove it from CI.

If there's a good reason to not remove Optimism, I can close and open PR with just the README changes.

All tests passed locally.


FlattestWhite (commented):
I think there are forked integration tests for optimism that should also be removed. As well as the .env.example file needs to be updated?

edkim:
Think you're right. I can update those.

edkim:
Will resubmit with only README changes

PR #133, opened 2023-02-22
Fix error in curve integration of ExchangeIssuanceLeveraged

ckoopmann:
Problem:
The main issue we face when using curve is the fact that they don't have a "exactOutput" swap. 
In the past we have used the CurveCalculator to determine the equivalent input amount from the requested output.
While this always had some minor imprecision, we were able to just add a few wei of "rounding" error on top to ensure successfull transactions.

Lately however the difference has grown by many orders of magnitudes (from a few wei to fractions of a percent of the total amount). 

Solution:
This smart contract change switches from using the CurveCalculator to just swapping all of the "maxInput" amount of tokens and then swapping the overbought amount back into the input token to achieve the effect of buying a fixed amount of output tokens. 
This is of course quite a bit more gas intensive but does bypass the whole issue with the CurveCalculator.

Open Issue:
-The "quote" method that is used by the frontend to get the required input amount has not been adjusted. There are a few ways around this:
a. Just add a small percentage of "slippage" on top of the quoted amount when issuing iceth
b. Static call the issue method using a tenderly fork or something from an impersonated address that has all of the funds necessary to execute the issuance
c. Adjust the quote method to maybe use an iterative method based on the `get_dx` function of the curve pool to approximate a better input estimation
(In this case gas costs are not really relevant since the quote method is only ever called statically)

### DIFF
--- a/contracts/exchangeIssuance/DEXAdapter.sol
+++ b/contracts/exchangeIssuance/DEXAdapter.sol
@@ -429,6 +429,9 @@ library DEXAdapter {
         );
         require(amountIn <= _maxAmountIn, "ExchangeIssuance: CURVE_OVERSPENT");
 
+        require(ICurvePool(_pool).get_dy(i, j, amountIn) >= _amountOut, "ExchangeIssuance: amountIn too low");
+
+
         if(_path[0] == ETH_ADDRESS){
             IWETH(_addresses.weth).withdraw(amountIn);
         }
--- a/contracts/exchangeIssuance/DEXAdapter.sol
+++ b/contracts/exchangeIssuance/DEXAdapter.sol
@@ -437,7 +437,7 @@ library DEXAdapter {
         }
 
         if(_path[_path.length-1] == ETH_ADDRESS){
-            IWETH(_addresses.weth).deposit{ value: returnedAmountOut }();
+            IWETH(_addresses.weth).deposit{ value: _amountOut }();
         }
 
         return _maxAmountIn.sub(swappedBackAmountIn);
--- a/contracts/exchangeIssuance/DEXAdapter.sol
+++ b/contracts/exchangeIssuance/DEXAdapter.sol
@@ -420,30 +420,27 @@ library DEXAdapter {
         require(_path.length == 2, "ExchangeIssuance: CURVE_WRONG_PATH_LENGTH");
         (int128 i, int128 j) = _getCoinIndices(_pool, _path[0], _path[1], ICurveAddressProvider(_addresses.curveAddressProvider));
 
-        uint256 amountIn = _getAmountInCurve(
-            _pool,
-            i,
-            j,
-            _amountOut,
-            _addresses
-        );
-        require(amountIn <= _maxAmountIn, "ExchangeIssuance: CURVE_OVERSPENT");
-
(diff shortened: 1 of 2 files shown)


contracts/exchangeIssuance/DEXAdapter.sol
```diff
@@ -420,27 +420,27 @@ library DEXAdapter {
...
+            IWETH(_addresses.weth).withdraw(_maxAmountIn);
         }
 
-        uint256 returnedAmountOut = _exchangeCurve(i, j, _pool, amountIn, _amountOut, _path[0]);
+        uint256 returnedAmountOut = _exchangeCurve(i, j, _pool, _maxAmountIn, _amountOut, _path[0]);
         require(_amountOut <= returnedAmountOut, "ExchangeIssuance: CURVE_UNDERBOUGHT");
 
+        uint256 swappedBackAmountIn;
+        if(returnedAmountOut > _amountOut){
+            swappedBackAmountIn = _exchangeCurve(j, i, _pool, returnedAmountOut.sub(_amountOut), 0, _path[1]);
```
ckoopmann, line 433: Can we get wrecked here because we have minimumOutput = 0 ? 
I expect not since we are doing a reverse swap of a bigger swap we did before so sandwiching / frontrunning should not be possible in my limited understanding, but we should think this through.

contracts/exchangeIssuance/DEXAdapter.sol
```diff
@@ -420,27 +420,27 @@ library DEXAdapter {
...
-            IWETH(_addresses.weth).withdraw(amountIn);
+            IWETH(_addresses.weth).withdraw(_maxAmountIn);
         }
 
-        uint256 returnedAmountOut = _exchangeCurve(i, j, _pool, amountIn, _amountOut, _path[0]);
+        uint256 returnedAmountOut = _exchangeCurve(i, j, _pool, _maxAmountIn, _amountOut, _path[0]);
         require(_amountOut <= returnedAmountOut, "ExchangeIssuance: CURVE_UNDERBOUGHT");
 
+        uint256 swappedBackAmountIn;
+        if(returnedAmountOut > _amountOut){
```
ckoopmann, line 432: Should we add a threshold here, so we don't waste gas on a swap of a few wei ? 
If so, should it be a fixed threshold or user provided ? (The latter while being more flexible would break backwards compatibility in the interface)

ElmoArmy: I don't have a problem with the behavior here. My reasoning being that if we go with a fixed threshold, we wouldn't have a way of picking a good number without pretty intricate testing. If we go with allowing the user to decide, we break the interface which slows down the application of this fix. We can optimize for gas later.

contracts/exchangeIssuance/DEXAdapter.sol
ElmoArmy: I thought through it as well and I agree with your initial instinct. That is to say that there is no way for a transaction to run between the two swaps and that no additional risk occurs by having the minimum out amount be 0. 

Alongside the introduction of a future feature that implements a minimum threshold to send dust back to the user, we could perhaps come up with a more reasonable minimum out from the pool.

ElmoArmy (approved):
Ideally this library would have it's own test suite but that is okay because the contracts that use it are well tested. LGTM.

FlattestWhite (approved):
LGTM

PR #141, opened 2023-06-01
Zero ex wrapper contract

ckoopmann:
This is a wrapper contract for use in our frontend selling / buying setTokens directly on 0x. (not relevant for FlashMinting).

This came up as a workaround in the context of the LedgerLive integration which requires us to decode any calldata we send using predefined method signatures. (which we cannot really specify when calling 0x directly due to the Router being an upgradeable diamond proxy).

As an additional benefit this contract ensures that the 0x router abides by the "min"/"max" amount restrictions.


ckoopmann:
Was wondering just now if we need the hardcoded callTarget address. Security wise we could probably let the user specify it, since we are already calling the user specified token addresses anyway.

Also this contract should never hold any token balance in the course of it's operation, so as long as there are no delegatecalls (which could be used to exploit user approvals), there should be no funds at risk. 

Would love to hear what @pblivin0x thinks about this.

ckoopmann:
Also was wondering if it would be worth it to extend this contract to allow trading with "permit" messages which could improve UX and security by removing the approval step.

contracts/periphery/ZeroExTradeWrapper.sol
```diff
@@ -0,0 +1,81 @@
...
+
+
+
+
+/**
+ * @title ZeroExTradeWrapper
+ * @author Index Coop
+ *
+ * Contract for wrapping ZeroEx trades
+ * Developed to fulfil calldata decoding requirments in the ledger integration
```
pblivin0x, line 26: nit: typos - fulfill, requirements

contracts/periphery/ZeroExTradeWrapper.sol
```diff
@@ -0,0 +1,81 @@
...
+            uint256 tokenOutBalance = _tokenOut.balanceOf(address(this));
+            require(tokenOutBalance >= _minAmountOut, "ZeroExTradeWrapper: Insufficient tokens received");
+            _tokenOut.transfer(msg.sender, tokenOutBalance);
+        } else {
+            require(address(this).balance >= _minAmountOut, "ZeroExTradeWrapper: Insufficient ETH received");
+        }
+
+        if (address(this).balance > 0) {
+            payable(msg.sender).transfer(address(this).balance);
+        }
```
pblivin0x, line 77: are these separate cases or can they be combined?

contracts/periphery/ZeroExTradeWrapper.sol
pblivin0x: let's also replace this description, since we have a function that moves funds it is for more than decoding

pblivin0x:
> Also was wondering if it would be worth it to extend this contract to allow trading with "permit" messages which could improve UX and security by removing the approval step.

i'm open to this, we should definitely experiment with Permit2 eventually. noting that this may add extra front end reqs and some inconsistencies

pblivin0x:
> Was wondering just now if we need the hardcoded callTarget address. Security wise we could probably let the user specify it, since we are already calling the user specified token addresses anyway.
> 
> Also this contract should never hold any token balance in the course of it's operation, so as long as there are no delegatecalls (which could be used to exploit user approvals), there should be no funds at risk.
> 
> Would love to hear what @pblivin0x thinks about this.

I agree on the user funds movement. Is there any clear benefit to not hardcoding callTarget at this time?

ckoopmann:
> > Was wondering just now if we need the hardcoded callTarget address. Security wise we could probably let the user specify it, since we are already calling the user specified token addresses anyway.
> > Also this contract should never hold any token balance in the course of it's operation, so as long as there are no delegatecalls (which could be used to exploit user approvals), there should be no funds at risk.
> > Would love to hear what @pblivin0x thinks about this.
> 
> I agree on the user funds movement. Is there any clear benefit to not hardcoding callTarget at this time?

Yes, the benefit would be that we don't have to redeploy the contract if 0x decides to change the router deployment / architecture.

ckoopmann:
> > Also was wondering if it would be worth it to extend this contract to allow trading with "permit" messages which could improve UX and security by removing the approval step.
> 
> i'm open to this, we should definitely experiment with Permit2 eventually. noting that this may add extra front end reqs and some inconsistencies

Yes, definetely beyond the scope of this PR but something we should keep in mind.

contracts/periphery/ZeroExTradeWrapper.sol
ckoopmann: Well this really is the main motivation why we are deploying this contract.  The 0x router already moves the funds, so this is not really an additional benefit that we get from using this wrapper contract over calling the 0x router directly.

The only additional benefit we get is enforcement of the tokenlimits instead of just trusting 0x to abide by it. I will add this to the description.

contracts/periphery/ZeroExTradeWrapper.sol
ckoopmann: We could remove the if check if that is what you mean. That would mean that in many cases we send a zero value transfer to the user address though, which I'd expect to be more gas intensive than the balance check. (Didn't actually evaluate this thoroughly though).

contracts/adapters/AaveV3LeverageStrategyExtension.sol
```diff
@@ -27,7 +27,7 @@ import {IBaseManager} from "../interfaces/IBaseManager.sol";
  * @title AaveV3LeverageStrategyExtension
  * @author Index Coop
  *
- * Extension of AaveLeverageStrategyExtension to add endpoint for setting the eMode categoryId
+ * Expanded version of the AaveLeverageStrategyExtension to add an endpoint for setting the eMode categoryId on AaveV3
```
pblivin0x, line 30: diff belongs in a different PR?

pblivin0x (changes requested):
requesting full test coverage for the ZeroExTradeWrapper

contracts/periphery/ZeroExTradeWrapper.sol
```diff
@@ -0,0 +1,88 @@
...
+                uint256 tokenOutBalance = _tokenOut.balanceOf(address(this));
+                require(tokenOutBalance >= _minAmountOut, "ZeroExTradeWrapper: Insufficient tokens received");
+                _tokenOut.transfer(msg.sender, tokenOutBalance);
+            } else {
+                require(address(this).balance >= _minAmountOut, "ZeroExTradeWrapper: Insufficient ETH received");
+            }
+        }
+
+        if (address(this).balance > 0) {
+            payable(msg.sender).transfer(address(this).balance);
```
pblivin0x, line 83: add test coverage

contracts/periphery/ZeroExTradeWrapper.sol
```diff
@@ -0,0 +1,88 @@
...
+            _tokenIn.transfer(msg.sender, _tokenIn.balanceOf(address(this)));
+        }
+
+        if(address(_tokenOut) != address(0)) {
+            if(address(_tokenOut) != ETH_ADDRESS) {
+                uint256 tokenOutBalance = _tokenOut.balanceOf(address(this));
+                require(tokenOutBalance >= _minAmountOut, "ZeroExTradeWrapper: Insufficient tokens received");
+                _tokenOut.transfer(msg.sender, tokenOutBalance);
+            } else {
+                require(address(this).balance >= _minAmountOut, "ZeroExTradeWrapper: Insufficient ETH received");
```
pblivin0x, line 78: add test coverage

PR #144, opened 2023-07-28
feat(extension): Add AuctionRebalanceExtension

pblivin0x:


contracts/adapters/AuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,197 @@
...
+        external
+        onlyOperator
+    {
+        address[] memory currentComponents = setToken.getComponents();
+        
+        require(currentComponents.length == _oldComponents.length, "Old components length must match the current components length.");
+
+        for (uint256 i = 0; i < _oldComponents.length; i++) {
+            require(currentComponents[i] == _oldComponents[i], "Input old components array must match the current components array.");
+        }
```
ElmoArmy, line 99: this logic is performed on the module.  Could you explain why it would be needed again here?

external/contracts/set/AuctionRebalanceModuleV1.sol
ElmoArmy, line 1: Did you want us to review the module in this PR? I guess it would make the tests easier, but wasn't sure if you wanted us to take a look at it here in a context outside of a mock.

external/contracts/set/ConstantPriceAdapter.sol
```diff
@@ -0,0 +1,74 @@
...
+pragma solidity 0.8.17;
+
+/**
+ * @title ConstantPriceAdapter
+ * @author Index Coop
+ * @notice Price adapter contract for AuctionRebalanceModuleV1 that returns a constant price.
+ * The rate of change is zero.
+ * Price formula: price = initialPrice
+ */
+contract ConstantPriceAdapter {
```
ElmoArmy, line 11: Might be a good idea to breakout the interface that the module expects and have this inherit/implement that interface. Mostly for completeness so somewhat optional, but if the idea is that we make more adapters... would be good to start on the right foot.

external/contracts/set/ConstantPriceAdapter.sol
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.17;
```
ElmoArmy, line 2: so hot🔥

ElmoArmy (commented):
started reviewing, want to run through the tests before approving but looks good at a glance.

contracts/adapters/AuctionRebalanceExtension.sol
pblivin0x: for the `currentComponents`, only the length of auction params provided is checked

https://github.com/IndexCoop/index-protocol/blob/9a7a951d1246b46f7ac28d453d35c6804e1d89ed/contracts/protocol/modules/v1/AuctionRebalanceModuleV1.sol#L723

https://github.com/IndexCoop/index-protocol/blob/9a7a951d1246b46f7ac28d453d35c6804e1d89ed/contracts/protocol/modules/v1/AuctionRebalanceModuleV1.sol#L1270-L1274

there is no check that the order of those auction params corresponds to the `SetToken.components()` order which will be used on the module. the extension performs this check

external/contracts/set/AuctionRebalanceModuleV1.sol
pblivin0x: no sir, only the extension and its call passing for this PR 🙌

external/contracts/set/ConstantPriceAdapter.sol
pblivin0x: leaving for now since i only added for unit testing, will hopefully setup better in a new repo 👀

FlattestWhite (approved):

snake-poison (approved):
LGTM!

PR #152, opened 2023-10-01
[mlp] Update flash mint contracts to use ISwapRouter02 interface

FlattestWhite:


test/integration/ethereum/addresses.ts
```diff
@@ -45,7 +45,7 @@ export const PRODUCTION_ADDRESSES = {
       router: "0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D",
     },
     uniV3: {
-      router: "0xE592427A0AEce92De3Edee1F18E0157C05861564",
+      router: "0x68b3465833fb72A70ecDF485E0e4C7bD8665Fc45",
```
ckoopmann, line 48: We should double check that this address is not used elsewhere, where we still need to use the old router version.

ckoopmann (approved):
LGTM. Changing out the existing implementation obviously means that we wont be able to use the old router interface anywhere anymore.

Afaik uniswap has deployed the SwapRouter02 everywhere though, so we should be fine. 
We need to remember to also adjust the address in the deployments repo though, otherwise we will be creating DexAdapter deployments that don't work.

PR #154, opened 2023-10-31
Global optimistic auction rebalance extension

snake-poison:
UMA docs: https://docs.uma.xyz/

Interaction Flow closely resembles:  https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/optimistic-governor/implementation/OptimisticGovernor.sol

### DIFF
--- a/contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
+++ b/contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
@@ -50,6 +50,33 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
     //     address indexed _setToken,
     //     address indexed _delegatedManager
     // );
+    event ProductSettingsUpdated(
+        IERC20 indexed setToken,
+        address indexed delegatedManager,
+        OptimisticRebalanceParams optimisticParams,
+        bytes32 indexed rulesHash
+    );
+    event RebalanceProposed(
+        ISetToken indexed setToken,
+        IERC20 indexed quoteAsset,
+        address[] oldComponents,
+        address[] newComponents,
+        AuctionExecutionParams[] newComponentsAuctionParams,
+        AuctionExecutionParams[] oldComponentsAuctionParams,
+        bool shouldLockSetToken,
+        uint256 rebalanceDuration,
+        uint256 positionMultiplier
+    );
+
+    event AssertedClaim(
+        IERC20 indexed _setToken,
+        address indexed _assertedBy,
+        bytes32 indexed rulesHash,
+        bytes32 _assertionId,
+        bytes _claimData
+    );
+
+
 
 
     /* ============ Structs ============ */
@@ -60,25 +87,11 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
         }
 
         struct OptimisticRebalanceParams{
-            address finder;            // Contract that finds UMA contracts on-chain.
             IERC20 collateral;        // Collateral currency used to assert proposed transactions.
             uint64  liveness;           // The amount of time to dispute proposed transactions before they can be executed.
             uint256 bondAmount;        // Configured amount of collateral currency to make assertions for proposed transactions.
             bytes32 identifier;        // Identifier used to request price from the DVM.
             OptimisticOracleV3Interface optimisticOracleV3; // Optimistic Oracle V3 contract used to assert proposed transactions.
-        }
-
-        struct RebalanceProposal{
-            uint256 proposeTime;    // Timestamp of when the proposal was proposed.
-            address _setToken;      // Address of the SetToken being rebalanced.
-            address _quoteAsset;    // Address of the quote asset used in the rebalance.
-            address[] _oldComponents; // Addresses of existing components in the SetToken.
-            address[] _newComponents; // Addresses of new components to be added.
-            AuctionExecutionParams[]  _newComponentsAuctionParams; // AuctionExecutionParams for new components, indexed corresponding to _newComponents.
-            AuctionExecutionParams[]  _oldComponentsAuctionParams; // AuctionExecutionParams for existing components, indexed corresponding to the current component positions. Set to 0 for components being removed.
-            bool _shouldLockSetToken; // Indicates if the rebalance should lock the SetToken.
-            uint256 _rebalanceDuration; // Duration of the rebalance in seconds.
-            uint256 _positionMultiplier; // Position multiplier at the time target units were calculated.
         }
 
         struct ProductSettings{
@@ -88,9 +101,7 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
 
     /* ============ State Variables ============ */
     
-    // IAuctionRebalanceModuleV1 public immutable auctionModule;  // AuctionRebalanceModuleV1
-
-    mapping (ISetToken=>ProductSettings) public productSettings; // Mapping of quote asset to ProductSettings
+    mapping (ISetToken=>ProductSettings) public productSettings; // Mapping of set token to ProductSettings
     mapping(bytes32 => bytes32) public assertionIds; // Maps proposal hashes to assertionIds.
     mapping(bytes32 => bytes32) public proposalHashes; // Maps assertionIds to proposal hashes.
     mapping (bytes32=>ISetToken) public assertedProducts; // Maps assertionIds to ISetToken
@@ -109,7 +120,7 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
     /* ============ External Functions ============ */
 
     /**
-    * @dev OPERATOR ONLY: sets product settings for a given quote asset
+    * @dev OPERATOR ONLY: sets product settings for a given set token
     * @param _product Address of the SetToken to set rules and settings for.
     * @param _optimisticParams OptimisticRebalanceParams struct containing optimistic rebalance parameters.
     * @param _rulesHash bytes32 containing the ipfs hash rules for the product.
@@ -126,10 +137,10 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
             optimisticParams: _optimisticParams,
             rulesHash: _rulesHash
         });
+     emit ProductSettingsUpdated(_product, _product.manager(), _optimisticParams, _rulesHash);
     }
 
      /**
-    * @dev OPERATOR ONLY: sets product settings for a given quote asset
      * @param _quoteAsset                   ERC20 token used as the quote asset in auctions.
      * @param _oldComponents                Addresses of existing components in the SetToken.
      * @param _newComponents                Addresses of new components to be added.
@@ -173,8 +184,6 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
 
         bytes memory claim = _constructClaim(proposalHash, settings.rulesHash);
 
-  
-
         
         uint256 minimumBond = settings.optimisticParams.optimisticOracleV3.getMinimumBond(address(settings.optimisticParams.collateral));
         uint256 totalBond =  minimumBond > settings.optimisticParams.bondAmount ? minimumBond : settings.optimisticParams.bondAmount;
@@ -189,14 +198,17 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
             address(0),
             settings.optimisticParams.liveness,
             settings.optimisticParams.collateral,
-            settings.optimisticParams.bondAmount,
+            totalBond,
             settings.optimisticParams.identifier,
             bytes32(0)
         );
+        // emit RebalanceProposed( _setToken, _quoteAsset, _oldComponents, _newComponents, _newComponentsAuctionParams, _oldComponentsAuctionParams, _shouldLockSetToken, _rebalanceDuration, _positionMultiplier);
+        // emit AssertedClaim(_setToken, msg.sender, settings.rulesHash, assertionId, claim);
+        // emit AssertedClaim(_set, address(_setToken), msg.sender, assertionId, claim);
 
         assertionIds[proposalHash] = assertionId;
         proposalHashes[assertionId] = proposalHash;
-
+        assertedProducts[assertionId] = _setToken;
     }
    
     /**
@@ -246,6 +258,7 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
 
         delete assertionIds[proposalHash];
         delete proposalHashes[assertionId];
+        delete assertedProducts[assertionId];
 
         ProductSettings memory settings = productSettings[_setToken];
         settings.optimisticParams.optimisticOracleV3.settleAndGetAssertionResult(assertionId);
@@ -312,6 +325,7 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
             // Delete the disputed proposal and associated assertionId.
             delete assertionIds[proposalHash];
             delete proposalHashes[assertionId];
+            delete assertedProducts[assertionId];
 
         } else {
             // If the sender is not the expected Optimistic Oracle V3, check if the expected Oracle has the assertion and if not delete.
--- a/contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
+++ b/contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
@@ -46,10 +46,6 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
 
     /* ============ Events ============ */
 
-    // event AuctionRebalanceExtensionInitialized(
-    //     address indexed _setToken,
-    //     address indexed _delegatedManager
-    // );
     event ProductSettingsUpdated(
         IERC20 indexed setToken,
         address indexed delegatedManager,
@@ -76,9 +72,7 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
         bytes _claimData
     );
 
-
-
-
+    event ProposalDeleted(bytes32 assertionID, Proposal indexed proposal);
     /* ============ Structs ============ */
 
         struct AuctionExtensionParams {
@@ -99,12 +93,16 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
             bytes32 rulesHash;
         }
 
+        struct Proposal{
+            bytes32 proposalHash;
+            ISetToken product;
+        }
+
     /* ============ State Variables ============ */
     
     mapping (ISetToken=>ProductSettings) public productSettings; // Mapping of set token to ProductSettings
     mapping(bytes32 => bytes32) public assertionIds; // Maps proposal hashes to assertionIds.
-    mapping(bytes32 => bytes32) public proposalHashes; // Maps assertionIds to proposal hashes.
-    mapping (bytes32=>ISetToken) public assertedProducts; // Maps assertionIds to ISetToken
+    mapping(bytes32 => Proposal) public proposedProduct; // Maps assertionIds to a Proposal.
 
     // Keys for assertion claim data.
     bytes public constant PROPOSAL_HASH_KEY = "proposalHash";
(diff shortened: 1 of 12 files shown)


contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -56,11 +61,11 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
 
         struct OptimisticRebalanceParams{
             address finder;            // Contract that finds UMA contracts on-chain.
```
snake-poison, line 63: Don't think we really need this guy.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -56,11 +61,11 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
...
-            address collateral;        // Collateral currency used to assert proposed transactions.
+            IERC20 collateral;        // Collateral currency used to assert proposed transactions.
             uint64  liveness;           // The amount of time to dispute proposed transactions before they can be executed.
             uint256 bondAmount;        // Configured amount of collateral currency to make assertions for proposed transactions.
             bytes32 identifier;        // Identifier used to request price from the DVM.
-            address optimisticOracleV3; // Optimistic Oracle V3 contract used to assert proposed transactions.
+            OptimisticOracleV3Interface optimisticOracleV3; // Optimistic Oracle V3 contract used to assert proposed transactions.
         }
 
         struct RebalanceProposal{
```
snake-poison, line 71: Don't think we will be needing this at all.

snake-poison (commented):
Should add some internal functions that would DRY the contract a bit

Should also add some events.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+import {OptimisticOracleV3Interface} from "../interfaces/OptimisticOracleV3Interface.sol";
+
+/**
+ * @title GlobalOptimisticAuctionRebalanceExtension
+ * @author Index Coop
+ *
+ * @dev Extension contract for interacting with the AuctionRebalanceModuleV1. This contract acts as a pass-through and functions 
+ * are only callable by operator. 
+ */
+contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExtension {
```
pblivin0x, line 43: should we inherit `OptimisticOracleV3CallbackRecipientInterface` ?https://github.com/UMAprotocol/protocol/blob/cef7d8880a2168eeb30f54151a8d40b274fbd381/packages/core/contracts/optimistic-governor/implementation/OptimisticGovernor.sol#L28C43-L28C43

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+import { IDelegatedManager } from "../interfaces/IDelegatedManager.sol";
+import {GlobalAuctionRebalanceExtension} from "./GlobalAuctionRebalanceExtension.sol";
+import {AncillaryData } from "../lib/AncillaryData.sol";
+import {OptimisticOracleV3Interface} from "../interfaces/OptimisticOracleV3Interface.sol";
+
+/**
+ * @title GlobalOptimisticAuctionRebalanceExtension
+ * @author Index Coop
+ *
+ * @dev Extension contract for interacting with the AuctionRebalanceModuleV1. This contract acts as a pass-through and functions
```
pblivin0x, line 40: update natspec with optimistic specific message

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+        struct ProductSettings{
+            OptimisticRebalanceParams optimisticParams;
+            bytes32 rulesHash;
+        }
+
+    /* ============ State Variables ============ */
+    
+    // IAuctionRebalanceModuleV1 public immutable auctionModule;  // AuctionRebalanceModuleV1
+
+    mapping (ISetToken=>ProductSettings) public productSettings; // Mapping of quote asset to ProductSettings
```
pblivin0x, line 93:
```suggestion
    mapping (ISetToken=>ProductSettings) public productSettings; // Mapping of set token to ProductSettings
```

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+    /* ============ Constructor ============ */
+
+    constructor(AuctionExtensionParams memory _auctionParams, OptimisticRebalanceParams memory _optimisticParams) public GlobalAuctionRebalanceExtension(_auctionParams.managerCore, _auctionParams.auctionModule) {
+    
+    }
+
+    /* ============ External Functions ============ */
+
+    /**
+    * @dev OPERATOR ONLY: sets product settings for a given quote asset
```
pblivin0x, line 112:
```suggestion
    * @dev OPERATOR ONLY: sets product settings for a given set token
```

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+        bytes32 _rulesHash
+    )
+        external
+        onlyOperator(_product)
+    {
+        productSettings[_product] = ProductSettings({
+            optimisticParams: _optimisticParams,
+            rulesHash: _rulesHash
+        });
+    }
```
pblivin0x, line 129: event emission here?

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+
+     /**
+    * @dev OPERATOR ONLY: sets product settings for a given quote asset
```
pblivin0x, line 132:
```suggestion
    * @dev OPERATOR ONLY: sets product settings for a given set token
```

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+        bytes memory claim = _constructClaim(proposalHash, settings.rulesHash);
+
+  
+
+        
+        uint256 minimumBond = settings.optimisticParams.optimisticOracleV3.getMinimumBond(address(settings.optimisticParams.collateral));
+        uint256 totalBond =  minimumBond > settings.optimisticParams.bondAmount ? minimumBond : settings.optimisticParams.bondAmount;
+
+        settings.optimisticParams.collateral.safeTransferFrom(msg.sender, address(this), totalBond);
+        settings.optimisticParams.collateral.safeIncreaseAllowance(address(settings.optimisticParams.optimisticOracleV3), totalBond);
```
pblivin0x, line 183: lets clean this up with some internal functions

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+            settings.optimisticParams.liveness,
+            settings.optimisticParams.collateral,
+            settings.optimisticParams.bondAmount,
+            settings.optimisticParams.identifier,
+            bytes32(0)
+        );
+
+        assertionIds[proposalHash] = assertionId;
+        proposalHashes[assertionId] = proposalHash;
+
```
pblivin0x, line 199: event emission?

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+        ));
+
+        bytes32 assertionId = assertionIds[proposalHash];
+        require(assertionId != bytes32(0), "Proposal hash does not exist");
+
+        delete assertionIds[proposalHash];
+        delete proposalHashes[assertionId];
+
+        ProductSettings memory settings = productSettings[_setToken];
+        settings.optimisticParams.optimisticOracleV3.settleAndGetAssertionResult(assertionId);
```
pblivin0x, line 251: this will revert if the assertion result is invalid? can we add some comments and note in the function natspec?

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
ckoopmann: This description seems to be generally wrong as this function proposes the rebalance claim to the oracle. 
Also right now this function **CAN BE CALLED BY ANYONE** and **NOT LIMITED TO OPERATOR ONLY**.
(This might be inteded in which case we should remove the misleading claim from the docstring)

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+    function proposeRebalance(
+        ISetToken _setToken,
+        IERC20 _quoteAsset,
+        address[] memory _oldComponents,
+        address[] memory _newComponents,
+        AuctionExecutionParams[] memory _newComponentsAuctionParams,
+        AuctionExecutionParams[] memory _oldComponentsAuctionParams,
+        bool _shouldLockSetToken,
+        uint256 _rebalanceDuration,
+        uint256 _positionMultiplier
```
ckoopmann, line 152: If we want this to be called by anyone we might want to add an argument to pass the maximum bond that the caller is willing to put up. Right now I think technically if someone sends this transaction, we could frontrun him change the bond settings to something crazy and potentially lock up all of his tokens. (if the sender has high enough approval).
Obviously not something we are going to do but might be a good addition to give potential callers of this method extra control over their funds.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+        settings.optimisticParams.collateral.safeIncreaseAllowance(address(settings.optimisticParams.optimisticOracleV3), totalBond);
+
+        bytes32 assertionId = settings.optimisticParams.optimisticOracleV3.assertTruth(
+            claim,
+            msg.sender,
+            address(this),
+            address(0),
+            settings.optimisticParams.liveness,
+            settings.optimisticParams.collateral,
+            settings.optimisticParams.bondAmount,
```
ckoopmann, line 192: Shouldn't this be "totalBond ? (instead of reading from the setting object)

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -91,7 +108,96 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
...
+        uint256 totalBond =  minimumBond > settings.optimisticParams.bondAmount ? minimumBond : settings.optimisticParams.bondAmount;
+
+        settings.optimisticParams.collateral.safeTransferFrom(msg.sender, address(this), totalBond);
+        settings.optimisticParams.collateral.safeIncreaseAllowance(address(settings.optimisticParams.optimisticOracleV3), totalBond);
+
+        bytes32 assertionId = settings.optimisticParams.optimisticOracleV3.assertTruth(
+            claim,
+            msg.sender,
+            address(this),
+            address(0),
```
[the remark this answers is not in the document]
ckoopmann: Setting this to the zero address means using the default UMA escalation mechanism I guess ?
Yeah, making this configurable could increase flexibility but also gives whoever controls the settings potential arbitrary control no ?

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
ckoopmann: Also what happens if someone else settled this assertion already by manually calling the oracle. Is this possible ? If so can this brick this method by letting this call revert ?

ckoopmann: Also why are we ignoring the assertion result here ? Shouldn't we check that the assertion passed ?

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+        );
+
+        _invokeManager(_manager((_setToken)), address(auctionModule), callData);
+    }
+
+     // Constructs the claim that will be asserted at the Optimistic Oracle V3.
+    function _constructClaim(bytes32 proposalHash, bytes32 rulesHash) internal pure returns (bytes memory) {
+        return
+            abi.encodePacked(
+                AncillaryData.appendKeyValueBytes32("", PROPOSAL_HASH_KEY, proposalHash),
```
ckoopmann, line 280: Hashing the proposal like this, means that you cannot validate the assertion using this claim alone. (because you cannot deccode the original proposal arguments from the hash).

Do you think this could be a problem for UMAs dispute resolution process ?

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,328 @@
...
+        require(address(settings.optimisticParams.optimisticOracleV3) != address(0), "Invalid oracle address");
+
+
+        // If the sender is the Optimistic Oracle V3, delete the proposal and associated assertionId.
+        if (msg.sender == address(settings.optimisticParams.optimisticOracleV3)) {
+            // Delete the disputed proposal and associated assertionId.
+            delete assertionIds[proposalHash];
+            delete proposalHashes[assertionId];
+
+        } else {
```
ckoopmann, line 316: Can you explain this else block maybe. Who other than the oracle would call this callback ? And in what case / for what purpose ?

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
ckoopmann: I agree with @pblivin0x some extra event emissions would probably be nice. Probably each of the external state changing methods could use a "success" event emission at the end.

ckoopmann (commented):
Good stuff. 
Mostly added questions for my own understanding.
Also might be worth to review all the docstrings at the end to make sure they align with the function implementations.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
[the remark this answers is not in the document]
pblivin0x: Agree

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
```diff
@@ -121,9 +227,29 @@ contract GlobalOptimisticAuctionRebalanceExtension is  GlobalAuctionRebalanceExt
...
+            _shouldLockSetToken,
+            _rebalanceDuration,
+            _positionMultiplier
+        ));
+
+        bytes32 assertionId = assertionIds[proposalHash];
+        require(assertionId != bytes32(0), "Proposal hash does not exist");
+
+        delete assertionIds[proposalHash];
+        delete proposalHashes[assertionId];
```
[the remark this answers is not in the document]
pblivin0x: bump to resolve @snake-poison

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
pblivin0x: should have test scenarios for all these

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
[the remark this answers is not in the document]
snake-poison: > Setting this to the zero address means using the default UMA escalation mechanism I guess ? 
Yes that is correct.

> Yeah, making this configurable could increase flexibility but also gives whoever controls the settings potential arbitrary control no ?

It could be an immutable constant that would require replacing the extension to update to make it less attractive for abuse. At least that is what I meant by making this setting available.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
snake-poison: wouldn't be as simple as just implementing the callbacks and running integration tests, but I think it is more proper while the former would also be good enough for an extension. I'd hope to be able to just have uma as dependency but working with the different compilers could quickly evolve to a lot of rewriting and even more glue for this repo.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
snake-poison: @ckoopmann @pblivin0x 

https://github.com/UMAprotocol/protocol/blob/cef7d8880a2168eeb30f54151a8d40b274fbd381/packages/core/contracts/optimistic-governor/implementation/OptimisticGovernor.sol#L319-L321

This is heavily inspired by the Optimistic Governor contract. 

> this will revert if the assertion result is invalid?

If the assertion were invalid or disputed, the assertion [ above ](https://github.com/UMAprotocol/protocol/blob/cef7d8880a2168eeb30f54151a8d40b274fbd381/packages/core/contracts/optimistic-governor/implementation/OptimisticGovernor.sol#L313C8-L313C8) would have reverted the transaction as a callback would run on disputes to remove the assertionId from the mapping. 

> Also what happens if someone else settled this assertion already by manually calling the oracle. Is this possible ? If so can this brick this method by letting this call revert ?

This function would just get the result in that case and if it had been disputed, than this code would not be reached. Fetching the same result twice does not make this call revert. 

settleAndGetAssertionResult on the oo works as a settle if needed and get the result for and does not revert if called on an already settled assertionid. 

> Also why are we ignoring the assertion result here ? Shouldn't we check that the assertion passed ?

I explain this above and I agree that this reads rather unintuitively. Please take a look at my response, and if you need no further clarification, maybe suggest a comment or refactor that makes this more clear. I can think to maybe add an in-line comment on the assertion that mirrors the documentation on the [ OptimisticGovernor.sol](https://github.com/UMAprotocol/protocol/blob/cef7d8880a2168eeb30f54151a8d40b274fbd381/packages/core/contracts/optimistic-governor/implementation/OptimisticGovernor.sol#L309C9-L312C20)
```
        // This will reject the transaction if the proposal hash generated from the inputs does not have the associated
        // assertionId stored. This is possible when a) the transactions have not been proposed, b) transactions have
        // already been executed, c) the proposal was disputed or d) the proposal was deleted after Optimistic Oracle V3
        // upgrade.
```

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
snake-poison: Solid catch. Will get this updated.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
snake-poison: > Hashing the proposal like this, means that you cannot validate the assertion using this claim alone. (because you cannot deccode the original proposal arguments from the hash).

So long as we emit the settings in an event, this would be easily verifiable with access to the chain.
 
> Do you think this could be a problem for UMAs dispute resolution process ?

This is the route that UMA's team went with for OSnap proposals. A hash of all the transactions while this implementation goes with a hash of the auction settings so I'm thinking it shouldn't be an issue for a disputer to listen for proposals, grab the event, grab the rules and make a decision on disputing.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
snake-poison: > Can you explain this else block maybe. Who other than the oracle would call this callback ? And in what case / for what purpose ?

This is applicable in the event of an upgrade between proposal and starting a rebalance. 
> https://github.com/UMAprotocol/protocol/blob/cef7d8880a2168eeb30f54151a8d40b274fbd381/packages/core/contracts/optimistic-governor/implementation/OptimisticGovernor.sol#L359C1-L380C6

It is made much more explicit here through an internal function. Would you prefer that here too?

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
snake-poison: resolved by removing.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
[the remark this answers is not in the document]
snake-poison: Leaving the constant address(0) in place for now.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
snake-poison: Added

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
snake-poison: Included now.

contracts/global-extensions/GlobalOptimisticAuctionRebalanceExtension.sol
snake-poison: Not implemented for now as it is outside of the scope to limit the bad actions the operator could take.

ckoopmann (approved):
LGTM! Great Work 🚀

PR #156, opened 2023-12-07
feat(extension): Add OptimisticAuctionRebalanceExtension

ckoopmann:


contracts/base-extensions/BaseOptimisticRebalanceExtension.sol
```diff
@@ -0,0 +1,390 @@
...
+import {OptimisticOracleV3Interface} from "../interfaces/OptimisticOracleV3Interface.sol";
+
+/**
+ * @title BaseOptimisticAuctionRebalanceExtension
+ * @author Index Coop
+ *
+ * @dev The contract extends `BaseAuctionRebalanceExtension` by adding an optimistic oracle mechanism for validating rules on the proposing and executing of rebalances. 
+ * It allows setting product-specific parameters for optimistic rebalancing and includes callback functions for resolved or disputed assertions.
+ */
+contract BaseOptimisticAuctionRebalanceExtension is  BaseAuctionRebalanceExtension, AssetAllowList {
```
pblivin0x, line 43: for consistency with [prior implementation](https://github.com/IndexCoop/index-coop-smart-contracts/blob/master/contracts/adapters/AuctionRebalanceExtension.sol), lets rename to `OptimisticAuctionRebalanceExtension`

contracts/base-extensions/BaseOptimisticRebalanceExtension.sol
```diff
@@ -0,0 +1,390 @@
...
+     * @param _oldComponentsAuctionParams   AuctionExecutionParams for existing components, indexed corresponding to
+     *                                      the current component positions. Set to 0 for components being removed.
+     * @param _shouldLockSetToken           Indicates if the rebalance should lock the SetToken.
+     * @param _rebalanceDuration            Duration of the rebalance in seconds.
+     * @param _positionMultiplier           Position multiplier at the time target units were calculated.
+     */
+    function startRebalance(
+        IERC20 _quoteAsset,
+        address[] memory _oldComponents,
+        address[] memory _newComponents,
```
pblivin0x, line 265: let's add our optional asset allow list check on the inputs `_oldComponents` and `_newComponents` here

contracts/base-extensions/BaseOptimisticRebalanceExtension.sol
ckoopmann: sure

contracts/base-extensions/BaseOptimisticRebalanceExtension.sol
ckoopmann: I think only checking "_newComponents" should be enough though. I think we will want to be able to rebalance out of non-allowed assets. (For example when we change the set of allowed assets, or we receive a non-allowed asset as an airdrop etc.)

pblivin0x: oh yes i agree, checking `_newComponents` is sufficient

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,390 @@
...
+        address indexed _assertedBy,
+        bytes32 indexed rulesHash,
+        bytes32 _assertionId,
+        bytes _claimData
+    );
+
+    event ProposalDeleted(
+        bytes32 assertionID, 
+        Proposal indexed proposal
+        );
```
pblivin0x, line 77: may i request we lint this contract to make spacing consistent?

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,390 @@
...
+
+    /* ============ Structs ============ */
+
+        struct AuctionExtensionParams {
+            IBaseManager baseManager;     // Manager Contract of the set token for which to deploy this extension
+            IAuctionRebalanceModuleV1 auctionModule; // Contract that rebalances index sets via single-asset auctions
+            bool useAssetAllowlist;     // Bool indicating whether to use asset allow list
+            address[] allowedAssets;    // Array of allowed assets
```
pblivin0x, line 85: are we storing these twice with AssetAllowList inheritance?

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,390 @@
...
+        address[] memory _oldComponents,
+        address[] memory _newComponents,
+        AuctionExecutionParams[] memory _newComponentsAuctionParams,
+        AuctionExecutionParams[] memory _oldComponentsAuctionParams,
+        bool _shouldLockSetToken,
+        uint256 _rebalanceDuration,
+        uint256 _positionMultiplier
+    )
+        external
+        onlyAllowedAssets(_newComponents)
```
pblivin0x, line 200: should we move this modifier over to `startRebalance`? considering a staleness case where the asset allowlist is modified between `proposeRebalance` and `startRebalance` calls.

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
ckoopmann: Nope no deuplications, this is just a struct definition that we use to encode the arguments to be sent to the constructor.

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
ckoopmann: Yes makes sense. Moved it over accordingly.

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,390 @@
...
+        }
+
+        struct Proposal{
+            bytes32 proposalHash;   // Hash of the proposal.
+            ISetToken product;    // Address of the SetToken to set rules and settings for.
+        }
+
+    /* ============ State Variables ============ */
+    
+    mapping (ISetToken=>ProductSettings) public productSettings; // Mapping of set token to ProductSettings
```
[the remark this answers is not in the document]
ckoopmann: Yep, we already manage only one setToken using this storage field. Forgot to remove this unnecessary mapping though, which I just fixed.

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
ckoopmann: Manually fixed the indentation in this case. Any automatic linter you can recommend for solidity ?

test/adapters/optimisticAuctionRebalanceExtension.spec.ts
```diff
@@ -0,0 +1,757 @@
...
+}
+
+// Base58 decoding function (make sure you have a proper Base58 decoding function)
+function base58ToHexString(base58String: string) {
+  const bytes = base58.decode(base58String); // Decode base58 to a buffer
+  const hexString = bufferToHex(bytes.slice(2)); // Convert buffer to hex, excluding the first 2 bytes
+  return "0x" + hexString;
+}
+
+describe.only("OptimisticAuctionRebalanceExtension", () => {
```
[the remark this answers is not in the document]
ckoopmann: Done

test/adapters/optimisticAuctionRebalanceExtension.spec.ts
```diff
@@ -0,0 +1,757 @@
...
+        subjectAuctionRebalanceModule,
+          subjectUseAssetAllowlist,
+          allowedAssets,
+      );
+      extension = extension.connect(operator.wallet);
+      await baseManager.addAdapter(extension.address);
+      return extension;
+    }
+
+    it("should set the correct manager core", async () => {
```
pblivin0x, line 147: let's check these test descriptions

test/adapters/optimisticAuctionRebalanceExtension.spec.ts
```diff
@@ -0,0 +1,757 @@
...
+      const actualAuctionRebalanceModule = await auctionExtension.auctionModule();
+      expect(actualAuctionRebalanceModule).to.eq(subjectAuctionRebalanceModule);
+    });
+    it("should be able to initialize extension and module at the same time", async () => {
+      const auctionExtension = await subject();
+      await expect(
+        auctionExtension
+          .initialize(),
+      ).to.not.be.reverted;
+    });
```
pblivin0x, line 166: what is this test? https://github.com/IndexCoop/index-coop-smart-contracts/blob/4aa9028018a2ac533b60b55bf01c03185285616b/contracts/adapters/AuctionRebalanceExtension.sol#L187

test/adapters/optimisticAuctionRebalanceExtension.spec.ts
```diff
@@ -0,0 +1,757 @@
...
+
+    it("should revert if module is initialized and extension is not", async () => {
```
pblivin0x, line 168: ?

test/adapters/optimisticAuctionRebalanceExtension.spec.ts
```diff
@@ -0,0 +1,757 @@
...
+      setV2Setup.auctionModule.address,
+      useAssetAllowlist,
+      allowedAssets,
+    );
+    auctionRebalanceExtension = auctionRebalanceExtension.connect(operator.wallet);
+  });
+
+  addSnapshotBeforeRestoreAfterEach();
+
+  describe("#constructor", () => {
```
pblivin0x, line 122: let's simplify the constructor tests, see https://github.com/IndexCoop/index-coop-smart-contracts/blob/4aa9028018a2ac533b60b55bf01c03185285616b/test/adapters/auctionRebalanceExtension.spec.ts#L83

test/adapters/optimisticAuctionRebalanceExtension.spec.ts
```diff
@@ -0,0 +1,757 @@
...
+              subjectCaller = await getRandomAccount();
+            });
+
+            it("should revert", async () => {
+              await expect(subject()).to.be.revertedWith("Must be operator");
+            });
+          });
+        });
+
+        describe("#removeAdapter", () => {
```
pblivin0x, line 742: let's remove this test from this file since its a BaseManager function

test/adapters/optimisticAuctionRebalanceExtension.spec.ts
ckoopmann: Removed. (Had copied this case from the global version and "fixed" it to work with the base manager which made it a duplicate of the next case).

test/adapters/optimisticAuctionRebalanceExtension.spec.ts
ckoopmann: Done

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,387 @@
...
+import {OptimisticOracleV3Interface} from "../interfaces/OptimisticOracleV3Interface.sol";
+
+/**
+ * @title BaseOptimisticAuctionRebalanceExtension
+ * @author Index Coop
+ *
+ * @dev The contract extends `BaseAuctionRebalanceExtension` by adding an optimistic oracle mechanism for validating rules on the proposing and executing of rebalances. 
+ * It allows setting product-specific parameters for optimistic rebalancing and includes callback functions for resolved or disputed assertions.
+ */
+contract OptimisticAuctionRebalanceExtension is  AuctionRebalanceExtension, AssetAllowList {
```
pblivin0x, line 42: can we version this contract `OptimisticAuctionRebalanceExtensionV1` and add a note in the natspec about the unique features 
+ Optional asset whitelist
+ No Locking
+ Check for if rebalancing open

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,387 @@
...
+     * @param _rebalanceDuration            Duration of the rebalance in seconds.
+     * @param _positionMultiplier           Position multiplier at the time target units were calculated.
+    */
+    function proposeRebalance(
+        IERC20 _quoteAsset,
+        address[] memory _oldComponents,
+        address[] memory _newComponents,
+        AuctionExecutionParams[] memory _newComponentsAuctionParams,
+        AuctionExecutionParams[] memory _oldComponentsAuctionParams,
+        bool _shouldLockSetToken,
```
pblivin0x, line 195: can we remove this? make is such that all optimistic auctions proposed through this contract do not lock the SetToken?

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,387 @@
...
+     * ONLY OPERATOR: Toggle useAssetAllowlist on and off. When false asset allowlist is ignored
+     * when true it is enforced.
+     *
+     * @param _useAssetAllowlist           Bool indicating whether to use asset allow list
+     */
+    function updateUseAssetAllowlist(bool _useAssetAllowlist) external onlyOperator {
+        _updateUseAssetAllowlist(_useAssetAllowlist);
+    }
+
+
```
pblivin0x, line 157: can we add a boolean flag `isOpen` which can be flipped by the `operator` and is checked at the beginning of `proposeRebalance` and `startRebalance` calls?

this is our V1 solution to having a rebalance cadence of every 6 months. we may replace with timelocks in a later version

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
```diff
@@ -0,0 +1,387 @@
...
+        address[] memory _oldComponents,
+        address[] memory _newComponents,
+        AuctionExecutionParams[] memory _newComponentsAuctionParams,
+        AuctionExecutionParams[] memory _oldComponentsAuctionParams,
+        bool _shouldLockSetToken,
+        uint256 _rebalanceDuration,
+        uint256 _positionMultiplier
+    )
+        external
+        override
```
edkim, line 271: Do you need to have `onlyOperator` here?

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
ckoopmann: We can easily do that. What would be the advantage though?  Wouldn't that make it less flexible ? Is there no conceivable scenario where we would want to lock it ? (if so why did we have it in the global extension contract? 🤔)

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
ckoopmann: Not 100% sure. Ideally the UMA integration should allow us to have this open for anyone to call. 
However maybe we want to keep the training wheels on for now and make it onlyOperator ? (or `onlyAllowedCallers` which would allow us to set this to be callable by anyone later).

@pblivin0x What do you think ?

ckoopmann: Either way we should make it consistent with the docstring comment though of course, which currently states this to be "operator only".

ckoopmann: This being callable by anyone seems to have been on on purpose though with the global extensions as shown by [this test case](https://github.com/IndexCoop/index-coop-smart-contracts/blob/d5ab03004853f068528a1962d9d0b025dd8ef0e5/test/global-extensions/globalOptimisticAuctionRebalanceExtension.spec.ts#L476).

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
ckoopmann: Done. Note that I set it up so that it will be set to "false" every time the startRebalance function has been called, and both propose and startRebalance are gated by this. That means that we will have to flip it bag to "true" for every rebalance. Lmk if that is not what you had in mind.

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
ckoopmann: I removed the argument from the `proposeRebalance` method. I left it in the startRebalance method signature to keep compatibility with the interface, even though that argument is superflous.

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
ckoopmann: Done in: https://github.com/IndexCoop/index-coop-smart-contracts/pull/156/commits/77b876c0664495853d4aa22604624e32e732babd

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
ckoopmann: Note that this means to in effect that the extension would be "bricked" if the operator key is lost or set to address(0).

Alternatively I can remove that logic of resetting isOpen to "false" after each rebalance in which case we would have to do that in a manual transaction after each rebalance.

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
pblivin0x: this is fine to be public, we can update docstrings to reflect this

test/adapters/optimisticAuctionRebalanceExtensionV1.spec.ts
```diff
@@ -0,0 +1,729 @@
...
+
+  return hexStr;
+}
+
+// Base58 decoding function (make sure you have a proper Base58 decoding function)
+function base58ToHexString(base58String: string) {
+  const bytes = base58.decode(base58String); // Decode base58 to a buffer
+  const hexString = bufferToHex(bytes.slice(2)); // Convert buffer to hex, excluding the first 2 bytes
+  return "0x" + hexString;
+}
```
pblivin0x, line 47: let's move this to a utility since we also define it in global extension tests

https://github.com/IndexCoop/index-coop-smart-contracts/blob/ba40dae30c5f1b556f0cfbdcc5ce067c3cca3e25/test/global-extensions/globalOptimisticAuctionRebalanceExtension.spec.ts#L28-L44

contracts/adapters/OptimisticAuctionRebalanceExtension.sol
pblivin0x: addressing in #157

contracts/adapters/AuctionRebalanceExtension.sol
```diff
@@ -73,7 +73,7 @@ contract AuctionRebalanceExtension is BaseExtension {
      * @param _newComponentsAuctionParams   AuctionExecutionParams for new components, indexed corresponding to _newComponents.
      * @param _oldComponentsAuctionParams   AuctionExecutionParams for existing components, indexed corresponding to
      *                                      the current component positions. Set to 0 for components being removed.
-     * @param _shouldLockSetToken           Indicates if the rebalance should lock the SetToken.
+     * @param _shouldLockSetToken           Indicates if the rebalance should lock the SetToken. Has to be false in this version
```
pblivin0x, line 76:
```suggestion
     * @param _shouldLockSetToken           Indicates if the rebalance should lock the SetToken.
```

Omit locking restriction from `AuctionRebalanceExtension` natspec

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
```diff
@@ -0,0 +1,406 @@
...
+            _newComponents,
+            _newComponentsAuctionParams,
+            _oldComponentsAuctionParams,
+            false, // We don't allow locking the set token in this version
+            _rebalanceDuration,
+            _positionMultiplier
+        ));
+        require(assertionIds[proposalHash] == bytes32(0), "Proposal already exists");
+        require(productSettings.rulesHash != bytes32(""), "Rules not set");
+        require(address(productSettings.optimisticParams.optimisticOracleV3) != address(0), "Oracle not set");
```
pblivin0x, line 232: Let's add unit tests that trip these require statements

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
```diff
@@ -0,0 +1,406 @@
...
+        return
+            abi.encodePacked(
+                AncillaryData.appendKeyValueBytes32("", PROPOSAL_HASH_KEY, proposalHash),
+                ",",
+                RULES_KEY,
+                ":\"",
+                rulesHash,
+                "\""
+            );
+    }
```
pblivin0x, line 349: can we add some comments here? maybe link to UMA resource on claim construction?

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
```diff
@@ -0,0 +1,406 @@
...
+        );
+
+        assertionIds[proposalHash] = assertionId;
+        proposedProduct[assertionId] = Proposal({
+            proposalHash: proposalHash,
+            product: setToken
+        });
+
+        emit RebalanceProposed( setToken, _quoteAsset, _oldComponents, _newComponents, _newComponentsAuctionParams, _oldComponentsAuctionParams,  _rebalanceDuration, _positionMultiplier);
+        emit AssertedClaim(setToken, msg.sender, productSettings.rulesHash, assertionId, claim);
```
pblivin0x, line 257: Let's add unit tests for these event emissions

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
```diff
@@ -0,0 +1,406 @@
...
+            address(this),
+            address(0),
+            productSettings.optimisticParams.liveness,
+            productSettings.optimisticParams.collateral,
+            totalBond,
+            productSettings.optimisticParams.identifier,
+            bytes32(0)
+        );
+
+        assertionIds[proposalHash] = assertionId;
```
pblivin0x, line 250: Let's double check unit tests for these lines

contracts/lib/AssetAllowList.sol
```diff
@@ -0,0 +1,151 @@
...
+
+    SPDX-License-Identifier: Apache License, Version 2.0
+*/
+
+pragma solidity 0.6.10;
+import { AddressArrayUtils } from "./AddressArrayUtils.sol";
+
+/**
+ * @title ModuleBase
+ * @author Set Protocol
```
pblivin0x, line 24: Let's patch copy pasta and add unit tests for this contract

pblivin0x: added here [`9175f17`](https://github.com/IndexCoop/index-coop-smart-contracts/pull/157/commits/9175f176bdcfe60d87bff9ed24d047d9f1f28056)

test/integration/ethereum/optimisticAuctionRebalanceExtenisonV1.spec.ts
```diff
@@ -0,0 +1,599 @@
...
+      beforeEach(async () => {
+        await baseManager.addExtension(auctionRebalanceExtension.address);
+      });
+
+      context("when the product settings have been set", () => {
+        let productSettings: any;
+        let identifier: string;
+        let liveness: BigNumber;
+        beforeEach(async () => {
+          identifier = utils.formatBytes32String("TestIdentifier"); // TODO: Check how do we ensure that our identifier is supported on UMAs whitelist
```
pblivin0x, line 168: 👀

test/integration/ethereum/optimisticAuctionRebalanceExtenisonV1.spec.ts
```diff
@@ -0,0 +1,599 @@
...
+          await identifierWhitelist.addSupportedIdentifier(identifier);
```
pblivin0x, line 169: check with UMA

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
ckoopmann: Not sure what you'd like to test here exactly ?

test/adapters/optimisticAuctionRebalanceExtensionV1.spec.ts
pblivin0x: added here [e2d2340](https://github.com/IndexCoop/index-coop-smart-contracts/pull/157/commits/e2d2340da0c518e2d21c7e58ed3f33b7fb5c48de)

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
pblivin0x: do we test that the claim is constructed correctly? or the bond is pulled?

ckoopmann: Will add test for pulling the bond, which is trivial. 
For testing the claim construction, right now it'd mean to reimplement the logic in the javascript tests and then compare the values, not sure how much value that'd bring though.

ckoopmann: @pblivin0x Or do you have another "source of truth" of what a "correctly constructed" claim should look like, to test against ?🤔

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
ckoopmann: Done: https://github.com/IndexCoop/index-coop-smart-contracts/commit/4d6113f9a42203ee9e3f55ac8a6a93feb37302a4

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
ckoopmann: I couldn't really find any specific rules on claim construction on the uma documentation.
As far as I understand we can define our own conventions on the formatting / interpretation of claims.
The UMA examples from the docs are usually english sentences such as:
- "The sky is blue"
- "Argentinia won the 2022 Fifa World Cup"
- "The price of X Token was Y USD at time Z"

Ideally we should have a resource of our own where this formatting will be defined. (Or maybe even a frontend app in the future which shows the correct interpretation of those parameters).

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
ckoopmann: Done: https://github.com/IndexCoop/index-coop-smart-contracts/commit/c8a85859cc8c185b5238cc9928eb7c41efcb7e7f

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
ckoopmann: See also:
https://github.com/IndexCoop/index-coop-smart-contracts/pull/156#discussion_r1441232920

ckoopmann: Added tests for both pulling the bond tand verifying the claim construction (by reproducing the same logic in JS) here:
https://github.com/IndexCoop/index-coop-smart-contracts/pull/157/commits/49bddfa9e343930b377d7d1cd5e8d1a7c9b94e9e

contracts/adapters/AuctionRebalanceExtension.sol
ckoopmann: @pblivin0x Why should we remove this, it is true no ?

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
ckoopmann: Added a comment with a link tot he OptimisticGovernor implementation: https://github.com/IndexCoop/index-coop-smart-contracts/pull/156/commits/ced1ad5a55b5b2eb00b8642028d9ab9ce3fd25c0

contracts/adapters/AuctionRebalanceExtension.sol
pblivin0x: It is true that we do not allow `_shouldLockSetToken` to be `false` for the `OptimisticAuctionRebalanceExtension`, but this is not the case for the `AuctionRebalanceExtension`

test/integration/ethereum/optimisticAuctionRebalanceExtenisonV1.spec.ts
pblivin0x: The default `ASSERT_TRUTH`  allows settlement to happen in one of two ways:
+ Return the 1 value if the claim is true.
+ Return the 0 value if the claim is false or cannot be resolved.

Here are more details for the identifier:
https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-170.md

When using the `ASSERT_TRUTH` identifier with `assertTruth`, you input the identifier as below: 

`0x4153534552545f54525554480000000000000000000000000000000000000000`

contracts/adapters/OptimisticAuctionRebalanceExtensionV1.sol
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    struct ProductSettings{
+        OptimisticRebalanceParams optimisticParams;     // OptimisticRebalanceParams struct containing optimistic rebalance parameters.
+        bytes32 rulesHash;      // IPFS hash of the rules for the product.
+    }
+
+    /* ============ State Variables ============ */
+    
+    ProductSettings public productSettings; // Mapping of set token to ProductSettings
```
pblivin0x, line 107:
```suggestion
    ProductSettings public productSettings;
```
nit: no longer a mapping

pblivin0x (approved):
LGTM 💯

We may want to peek the flaky optimism integration test

ckoopmann:
> LGTM 💯
> 
> We may want to peek the flaky optimism integration test

Seems to be an issue on alchemy side for the optimism rpc. 
Let's keep an eye on it. Not much we can do about it though beyond considering an alternative rpc provider.

PR #169, opened 2024-04-11
FlashMintLeveraged: Support for fixed input Issuance

ckoopmann:


contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,181 @@
...
+        uint256 inputTokenBalanceAfter;
+        if( _inputToken == DEXAdapter.ETH_ADDRESS) {
+            inputTokenBalanceAfter = msg.sender.balance;
+        } else {
+            inputTokenBalanceAfter = IERC20(_inputToken).balanceOf(msg.sender);
+        }
+
+        return inputTokenBalanceBefore.sub(inputTokenBalanceAfter);
+    }
+
```
pblivin0x, line 179: like the idea of returning ETH here 👍

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,181 @@
...
+        uint256 inputTokenBalanceAfter;
+        if( _inputToken == DEXAdapter.ETH_ADDRESS) {
+            inputTokenBalanceAfter = msg.sender.balance;
+        } else {
+            inputTokenBalanceAfter = IERC20(_inputToken).balanceOf(msg.sender);
+        }
+
+        return inputTokenBalanceBefore.sub(inputTokenBalanceAfter);
+    }
```
pblivin0x, line 178: also like the idea of an iterative heuristic approach here 👍

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,590 @@
...
+import { IWETH } from "../interfaces/IWETH.sol";
+
+
+
+/**
+ * @title FlashMintLeveragedExtended
+ * @author Index Coop
+ *
+ * Extended version of FlashMintLeveraged which allows for exactInputIssuance and exactOutputRedemption
+ */
```
pblivin0x, line 37: can we flush out natspec for the contract and external functions?

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,590 @@
...
+        ISetToken _setToken,
+        uint256 _maxSetAmount,
+        address _outputToken,
+        uint256 _outputTokenAmount,
+        DEXAdapter.SwapData memory _swapDataCollateralForDebt,
+        DEXAdapter.SwapData memory _swapDataOutputToken,
+        DEXAdapter.SwapData memory _swapDataDebtForCollateral,
+        DEXAdapter.SwapData memory _swapDataInputToken,
+        uint256 _priceEstimateInflator,
+        uint256 _maxDust
```
pblivin0x, line 75: should these be contract level owner controlled params? or passed on every call by the user or in the background by the front end?

can we also add some extra natspec of these - what values will yield target UX on arbitrum?

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,590 @@
...
+            _maxSetAmount,
+            _outputToken,
+            _outputTokenAmount,
+            _swapDataDebtForCollateral,
+            _swapDataInputToken,
+            _priceEstimateInflator,
+            _maxDust,
+            outputTokenBalanceBefore
+        );
+        uint256 outputTokenObtained = IERC20(_outputToken).balanceOf(address(this)).sub(outputTokenBalanceBefore);
```
pblivin0x, line 101: should we also check `maxSetAmount` here?

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,590 @@
...
+        require(outputTokenObtained >= _outputTokenAmount, "FlashMintLeveragedExtended: outputTokenBalanceAfter < _outputTokenAmount");
```
pblivin0x, line 102: should this be an exact check?

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,590 @@
...
+        public
+        FlashMintLeveraged(_addresses, _setController, _debtIssuanceModule, _aaveLeverageModule, _aaveV3Pool, _vault)
+    {
+    }
+
+    function redeemSetForExactERC20(
+        ISetToken _setToken,
+        uint256 _maxSetAmount,
+        address _outputToken,
+        uint256 _outputTokenAmount,
```
pblivin0x, line 69: this variables may be more clearly named as `_exactOutputTokenAmount`

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,590 @@
...
+        address _vault
+    )
+        public
+        FlashMintLeveraged(_addresses, _setController, _debtIssuanceModule, _aaveLeverageModule, _aaveV3Pool, _vault)
+    {
+    }
+
+    function redeemSetForExactERC20(
+        ISetToken _setToken,
+        uint256 _maxSetAmount,
```
pblivin0x, line 67: noting that how `_maxSetAmount is shown on the front end and passed to the contract may need some fine tuning

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,590 @@
...
+        DEXAdapter.SwapData memory _swapDataInputToken,
+        uint256 _priceEstimateInflator,
+        uint256 _maxDust
+    )
+        internal
+        returns(uint256)
+    {
+        require(_inputTokenAmount > _maxDust, "FlashMintLeveragedExtended: _inputToken must be more than _maxDust");
+
+        while (_inputTokenAmount > _maxDust) {
```
pblivin0x, line 392: can we replace this while loop with a contract level parameter `MAX_ITERATIONS` check?

either way if the params of the call are proper then this value shouldn't be hit. lets be sure to add natspec clarifying

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,590 @@
...
+                _setToken,
+                _minSetAmount,
+                _inputToken,
+                _inputTokenAmount,
+                _swapDataDebtForCollateral,
+                _swapDataInputToken
+            );
+            _inputTokenAmount = _inputTokenAmount - inputTokenAmountSpent;
+            uint256 priceEstimate = _minSetAmount.mul(_priceEstimateInflator).div(inputTokenAmountSpent);
+            _minSetAmount = _inputTokenAmount.mul(priceEstimate).div(1 ether);
```
pblivin0x, line 403: do we want a `preciseMul` here?

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,590 @@
...
+        while (_inputTokenAmount > _maxDust) {
+            uint256 inputTokenAmountSpent = _initiateIssuanceAndReturnInputAmountSpent(
+                _setToken,
+                _minSetAmount,
+                _inputToken,
+                _inputTokenAmount,
+                _swapDataDebtForCollateral,
+                _swapDataInputToken
+            );
+            _inputTokenAmount = _inputTokenAmount - inputTokenAmountSpent;
```
pblivin0x, line 401: lets add some inline comments specifically for these 3 lines

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
```diff
@@ -0,0 +1,590 @@
...
+            uint256 priceEstimate = _minSetAmount.mul(_priceEstimateInflator).div(inputTokenAmountSpent);
+            _minSetAmount = _inputTokenAmount.mul(priceEstimate).div(1 ether);
+        }
+
+        // TODO: Decide what to do with the dust (maybe swap to eth and return as gas subsidy)
```
pblivin0x, line 406: i think its ideal to return dust in the form of ETH rebate here. we can also add a check that this ETH value is not too large

3 checks
- input token amount matches exactly
- set token amount is at least the minimum amount
- eth gas rebate does not exceed contract setting (if it did, that would mean misparameterization is likely)

test/integration/ethereum/flashMintLeveragedExtended.spec.ts
```diff
@@ -0,0 +1,706 @@
...
+  exchange: Exchange;
+};
+
+if (process.env.INTEGRATIONTEST) {
+  describe.only("FlashMintLeveragedExtended - Integration Test", async () => {
+    const addresses = process.env.USE_STAGING_ADDRESSES ? STAGING_ADDRESSES : PRODUCTION_ADDRESSES;
+    let owner: Account;
+    let deployer: DeployHelper;
+
+    let rEth: StandardTokenMock;
```
pblivin0x, line 44: this works for now but lets add some basic arbitrum integration tests in later PR's

contracts/exchangeIssuance/FlashMintLeveragedExtended.sol
pblivin0x: SGTM, will need to help front end integration set those params for Arbitrum

PR #177, opened 2024-06-17
Add EIP-712 signature verification

ckoopmann:


pblivin0x:
functionality added here https://github.com/IndexCoop/periphery/blob/main/src/staking/SignedSnapshotStakingPool.sol

PR #178, opened 2024-06-17
Migrate PrtStakingPool Contract to v0.8

ckoopmann:
- Add separate installation of solidity 0.8 compatible openzeppelin contracts (containing both EIP712 and ERC20Snapshots)
- Add script to remove duplicate type definitions from contracts used in both oz libraries

Note: Currently does not work and would require upgrading a few other dependencies (mainly ethers) which is a bit tricky.
Might be better to create a new repo for solidity 0.8 contracts with modern tooling.


pblivin0x:
Updated to v0.8 here https://github.com/IndexCoop/periphery/blob/main/src/staking/SnapshotStakingPool.sol

PR #182, opened 2024-08-19
Add `TargetWeightWrapExtension`

pblivin0x:


contracts/adapters/TargetWeightWrapExtension.sol
```diff
@@ -0,0 +1,382 @@
...
+        address _targetAsset,
+        uint256 _reserveUnits
+    )
+        external
+        nonReentrant
+        onlyAllowedRebalancer
+    {
+        require(isRebalancingActive, "Rebalancing is not active");
+        require(rebalanceInfo.targetAssets.contains(_targetAsset), "Invalid target asset");
+        require(isReserveOverweight(), "Reserve asset is not overweight");
```
edkim, line 142: Wouldn't it be better for to allow wraps so long as the reserve is _not underweight_? 

That way rebalancers can maximize performance be keeping less of the value in the non-yielding reserve asset.

If we require the reserve asset to be overweight, there may be situations where a yielding asset is under its target weight, but we can't rebalance into it.

contracts/adapters/TargetWeightWrapExtension.sol
```diff
@@ -0,0 +1,382 @@
...
+            _targetAsset,
+            _targetUnits,
+            executionParams[_targetAsset].wrapAdapterName,
+            executionParams[_targetAsset].unwrapData
+        );
+        invokeManager(address(wrapModule), data);
+
+        (uint256 targetAssetWeight, uint256 reserveWeight) = getTargetAssetAndReserveWeight(_targetAsset);
+        require(targetAssetWeight > executionParams[_targetAsset].minTargetWeight, "Target asset underweight post-unwrap");
+        require(reserveWeight < rebalanceInfo.maxReserveWeight, "Reserve asset overweight post-unwrap");
```
edkim, line 192: Noting that this function doesn't limit rebalancers from unwrapping "too much" target asset, and hurting yield performance. This is less of an issue if we allow wrapping when the reserve is below its max but above minimum. See suggestion above in wrap().

edkim (changes requested):
Great work! Super clean code. I only had one suggestion about being able to wrap into yielding assets so long as the reserve is above its Min weight.

contracts/adapters/TargetWeightWrapExtension.sol
```diff
@@ -0,0 +1,382 @@
...
+     * @notice Sets the reserve asset, target assets, and their associated execution parameters.
+     * @dev Only callable by the operator.
+     * @dev The weights are percentages where 100% equals 1e18.
+     * @param _reserveAsset Address of the reserve asset.
+     * @param _minReserveWeight Minimum allowable weight of the reserve asset.
+     * @param _maxReserveWeight Maximum allowable weight of the reserve asset.
+     * @param _targetAssets Array of target asset addresses.
+     * @param _executionParams Array of execution parameters corresponding to each target asset.
+     */
+    function setTargetWeights(
```
edkim, line 207: If we want to remove a component we would set the min and max weight to zero, right?

contracts/adapters/TargetWeightWrapExtension.sol
pblivin0x: yes this makes sense...in this case it may make sense to remove the pre-wrap weight check entirely

contracts/adapters/TargetWeightWrapExtension.sol
pblivin0x: good catch on extrema - yes i think it makes sense here, but the target asset weight checks in `wrap()` and `unwrap()` will need to be updated from `<, >` to `<=, >=`

contracts/adapters/TargetWeightWrapExtension.sol
pblivin0x: i agree, these functions make sense in conjunction with each other if the conditions are asymmetric and `wrap()` doesn't check reserve weight before, only after

contracts/adapters/TargetWeightWrapExtension.sol
pblivin0x: updated checks and added tests for removing both the component and reserve asset here [`4269d1e`](https://github.com/IndexCoop/index-coop-smart-contracts/pull/182/commits/4269d1ee222d2de41f3a5572ab5e0b2fa28aeda7)

contracts/adapters/TargetWeightWrapExtension.sol
pblivin0x: removed condition on `wrap()` here [`4269d1e`](https://github.com/IndexCoop/index-coop-smart-contracts/pull/182/commits/4269d1ee222d2de41f3a5572ab5e0b2fa28aeda7)

edkim (approved):
Changes LGTM!

contracts/adapters/TargetWeightWrapExtension.sol
```diff
@@ -0,0 +1,385 @@
...
+            _targetAsset,
+            _reserveUnits,
+            executionParams[_targetAsset].wrapAdapterName,
+            executionParams[_targetAsset].wrapData
+        );
+        invokeManager(address(wrapModule), data);
+
+        (uint256 targetAssetWeight, uint256 reserveWeight) = getTargetAssetAndReserveWeight(_targetAsset);
+        require(targetAssetWeight <= executionParams[_targetAsset].maxTargetWeight, "Target asset overweight post-wrap");
+        require(reserveWeight >= rebalanceInfo.minReserveWeight, "Reserve asset underweight post-wrap");
```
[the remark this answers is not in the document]
pblivin0x: In this use case most of the wrap/unwrap operations will have negligible slippage, so we're unlikely to have other target assets become overweight/underweight as a result of a different target asset being wrapped/unwrapped.

contracts/adapters/TargetWeightWrapExtension.sol
```diff
@@ -0,0 +1,385 @@
...
+        uint256 maxReserveWeight,
+        address[] targetAssets,
+        TargetWeightWrapParams[] executionParams
+    );
+    event RebalancePaused();
+
+    /* ========== Immutables ========= */
+
+    ISetToken public immutable setToken;
+    IWrapModuleV2 public immutable wrapModule;
```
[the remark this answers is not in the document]
pblivin0x: updated to `wrapModule` and `setValuer` mutable here [`666ecd2`](https://github.com/IndexCoop/index-coop-smart-contracts/pull/182/commits/666ecd29ed5cdf5726291f1fe10a072329cb0577)

contracts/adapters/TargetWeightWrapExtension.sol
```diff
@@ -0,0 +1,385 @@
...
+        uint256 _maxReserveWeight,
+        address[] memory _targetAssets,
+        TargetWeightWrapParams[] memory _executionParams
+    )
+        external
+        onlyOperator
+    {
+        require(_targetAssets.length == _executionParams.length, "Mismatched array lengths");
+        require(_minReserveWeight <= _maxReserveWeight, "Invalid min reserve weight");
+        require(_maxReserveWeight <= PreciseUnitMath.preciseUnit(), "Invalid max reserve weight");
```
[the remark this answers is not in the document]
pblivin0x: It's sometimes desirable to set individual targets such that the overall targets are either overweight or underweight to induce specific rebalancing behaviors. Similar to the `AuctionRebalanceModuleV1`, setting targets above the current value may increase the likelihood that the quote asset is spent or that the reserve approaches the minimum. Conversely, setting targets below the current value may encourage saving the quote asset, keeping the reserve closer to its maximum. Therefore, we'll continue with minimally opinionated checks on individual asset weights.

contracts/adapters/TargetWeightWrapExtension.sol
```diff
@@ -0,0 +1,385 @@
...
+            _targetAsset,
+            _targetUnits,
+            executionParams[_targetAsset].wrapAdapterName,
+            executionParams[_targetAsset].unwrapData
+        );
+        invokeManager(address(wrapModule), data);
+
+        (uint256 targetAssetWeight, uint256 reserveWeight) = getTargetAssetAndReserveWeight(_targetAsset);
+        require(targetAssetWeight >= executionParams[_targetAsset].minTargetWeight, "Target asset underweight post-unwrap");
+        require(reserveWeight <= rebalanceInfo.maxReserveWeight, "Reserve asset overweight post-unwrap");
```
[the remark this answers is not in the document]
pblivin0x: Added comment on intended wrap/unwrap operations being minimal slippage here [`cff5d06`](https://github.com/IndexCoop/index-coop-smart-contracts/pull/182/commits/cff5d06befd392b87591a8e23c5a40f5a572a25d)

contracts/adapters/TargetWeightWrapExtension.sol
```diff
@@ -0,0 +1,385 @@
...
+    ISetToken public immutable setToken;
+    IWrapModuleV2 public immutable wrapModule;
+    ISetValuer public immutable setValuer;
+
+    /* ========== State Variables ========= */
+
+    bool public isRebalancingActive;
+    bool public isRebalanceOpen;
+
+    mapping(address => TargetWeightWrapParams) public executionParams;
```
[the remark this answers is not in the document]
pblivin0x: updated documentation here [`1795554`](https://github.com/IndexCoop/index-coop-smart-contracts/pull/182/commits/1795554add12ddd853c80392d2f8b8ea29300794)

contracts/adapters/TargetWeightWrapExtension.sol
```diff
@@ -0,0 +1,385 @@
...
+
+    /* ========== Immutables ========= */
+
+    ISetToken public immutable setToken;
+    IWrapModuleV2 public immutable wrapModule;
+    ISetValuer public immutable setValuer;
+
+    /* ========== State Variables ========= */
+
+    bool public isRebalancingActive;
```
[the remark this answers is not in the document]
pblivin0x: updated documentation here [`1795554`](https://github.com/IndexCoop/index-coop-smart-contracts/pull/182/commits/1795554add12ddd853c80392d2f8b8ea29300794)

PR #184, opened 2024-09-03
Add `RebasingComponentModule` with `FlashMintWrapped` integration tests

pblivin0x:


edkim (approved):
LGTM. Could be helpful to add tests for issuing from / redeeming to another currency, but leave that up to you 😄

PR #188, opened 2024-10-11
WIP: DexAdapterV3 with BalancerV2 support

edkim:
To support multi-hop swaps on BalV2, I needed to add a `poolIds` array to the SwapData struct. 
I believe we need that for redemptions, with the swap path being agETH -> rsETH -> WETH through [this pool](https://balancer.fi/pools/ethereum/v2/0xf1bbc5d95cd5ae25af9916b8a193748572050eb00000000000000000000006bc) and then [this one](https://balancer.fi/pools/ethereum/v2/0x58aadfb1afac0ad7fca1148f3cde6aedf5236b6d00000000000000000000067f)

### DIFF
--- a/contracts/exchangeIssuance/FlashMintHyETHV3.sol
+++ b/contracts/exchangeIssuance/FlashMintHyETHV3.sol
@@ -375,7 +375,8 @@ contract FlashMintHyETHV3 is Ownable, ReentrancyGuard {
             // Special handling for agETH
             if (marketData.underlying == address(agETH)) {
                 rsEthAdapter.getRSETHWithETH{value: ethAmount}("");
-                marketData.sy.deposit(address(this), address(agETH), ethAmount, 0);
+                uint256 agEthAmount = agETH.balanceOf(address(this));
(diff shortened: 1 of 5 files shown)


edkim:
I haven't been able to get the FlashMintHyETH issuance tests to pass yet. Here's the `cast run` trace i get from my last commit https://gist.github.com/edkim/084c806e102c6d798f9e75fbf86832b4

contracts/exchangeIssuance/FlashMintHyETHV3.sol
```diff
@@ -378,7 +375,8 @@ contract FlashMintHyETHV3 is Ownable, ReentrancyGuard {
             // Special handling for agETH
             if (marketData.underlying == address(agETH)) {
                 rsEthAdapter.getRSETHWithETH{value: ethAmount}("");
-                marketData.sy.deposit(address(this), address(agETH), ethAmount, 0);
+                uint256 agEthAmount = agETH.balanceOf(address(this));
```
edkim, line 378: The amount of agETH returned from the rsETH adapter is different than `ethAmount`, so I get the balance and use that instead.

